PR #14, opened 2022-10-06
Adding and refactoring Unit tests for OME Tiff, OME Zarr, Open Slide Unit Tests

ktsitsi:
Extensive refactoring of all testing suite. Since there are many levels of modules which are used (wrapped etc..) we split the test suite into `unit` and `integration` testing. Some of the tests use real data stored locally in repo or s3, for trivial cases and functions unit tests we either mock or create random data.
```
---------- coverage: platform darwin, python 3.9.13-final-0 ----------
Name                                Stmts   Miss  Cover   Missing
-----------------------------------------------------------------
tiledbimg/__init__.py                   0      0   100%
tiledbimg/converters/__init__.py        0      0   100%
tiledbimg/converters/base.py           55      1    98%   64
tiledbimg/converters/ome_tiff.py       18      0   100%
tiledbimg/converters/ome_zarr.py       20      0   100%
tiledbimg/converters/openslide.py      20      1    95%   31
tiledbimg/openslide.py                 52      9    83%   24-25, 75-79, 83-84
-----------------------------------------------------------------
TOTAL                                 165     11    93%

```


tests/unit/converters/test_ome_zarr.py
```diff
@@ -0,0 +1,40 @@
...
+    def test_ome_zarr_level_downsamples(self, tmp_path):
+        zarr_path = os.path.join(tmp_path, "test.zarr")
+        reader = OMEZarrReader(zarr_path)
+        assert reader.level_downsamples == ()
+
+    def test_ome_zarr_level_image(self, tmp_path, mocker_tiff):
+        zarr_path = os.path.join(tmp_path, "test.zarr")
+        reader = OMEZarrReader(zarr_path)
+        zarr_array = np.array(
+            [[[[[[[0, 1], [2, 3]], [[4, 5], [6, 7, 8]]]]]]], dtype=object
```
gsakkis, line 33:
```
In [5]: zarr_array
Out[5]: 
array([[[[[[list([0, 1]), list([2, 3])],
           [list([4, 5]), list([6, 7, 8])]]]]]], dtype=object)
```
That's not realistic data, the Ome-Zarr images are 5D arrays and do not hold nested lists:
```
In [18]: z = zarr.open("tests/data/CMU-1-Small-Region.ome.zarr/")

In [19]: z[0][0]
Out[19]: <zarr.core.Array '/0/0' (1, 3, 1, 2967, 2220) uint8>
```

I am sceptical of the utility of unit tests (at least for this repo) and this test illustrates why.

tests/unit/converters/test_ome_zarr.py
```diff
@@ -0,0 +1,40 @@
...
+
+import numpy as np
+import pytest
+import zarr
+
+from tiledbimg.converters.ome_zarr import OMEZarrReader
+
+
+@pytest.fixture
+def mocker_tiff(mocker):
```
gsakkis, line 11: s/mocker_tiff/mocker_zarr

tests/unit/converters/test_ome_zarr.py
```diff
@@ -0,0 +1,40 @@
...
+    mocker.patch("zarr.open")
+
+
+class TestOMEZarrReader:
+    def test_ome_zarr_level_count(self, tmp_path):
```
gsakkis, line 16: This test and the next one don't use the mocked `zarr.open`

gsakkis (changes requested):
A few (mostly minor) comments on the integration tests. 

As for the newly added unit tests, I am sceptical of their utility, at least for this repo:
- they are often trivial (e.g. set a value to a mock attribute - assert that this value is written)
- the mocked values may be unrealistic (or completely bogus).
- they imply knowledge of (and depend on) the implementation internals (e.g. `reader._osd.level_downsamples = (5.6, 7.4, 6.7)`)

Overall they give me little (if any) confidence about the actual correctness of the code. I'd remove them keep just the integration tests (but don't object to them if others find them useful).

tests/integration/converters/__init__.py
```diff
@@ -0,0 +1,51 @@
+from dataclasses import dataclass
+
+import numpy as np
+import tiledb
+
+
+@dataclass(frozen=True)
+class CMU_1_SMALL_REGION:
```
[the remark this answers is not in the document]
ktsitsi: Why the representation of a dataset with minimal information and minimal functionality is better depicted as a function?

tests/integration/converters/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+    _dom = {
+        0: ((0, 2219), (0, 2966)),
+        1: ((0, 386), (0, 462)),
+        2: ((0, 1279), (0, 430)),
+    }
+    _attr_dtype = [("f0", "u1"), ("f1", "u1"), ("f2", "u1")]
+    _tile = {0: (1024, 1024), 1: (387, 463), 2: (1024, 431)}
```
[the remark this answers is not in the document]
ktsitsi: Why to sacrifice the self_explanatory nature of this naming and zip values so we can save one loc?

tests/unit/converters/test_ome_zarr.py
ktsitsi: IMO unit tests do not have to be realistic. On the realistic side at least to my mind are the integration (end-to-end) ones. Unit are there to ensure that the code actual output is the expected output according to input but no matter kind is the input. Of course since this repo pretty much wraps well known and tested functions like the ones from `numpy` etc... maybe it is unclear why we have them but in general testing parts of code even without real data is sth legit.

tests/unit/converters/test_ome_zarr.py
ktsitsi: Yes I will add although it does not make a change since `zarr.open` accepts empty path as input, but for completeness I will add it

tests/unit/test_openslide.py
```diff
@@ -0,0 +1,68 @@
+from __future__ import annotations
+
+import os
+
+import openslide as osld
+import tiledb
+
+from tests import download_from_s3
+from tests.integration.converters import CMU_1_SMALL_REGION
```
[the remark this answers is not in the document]
ktsitsi: `CMU_1_SMAL_REGION` dataset can be easily moved to `__init__` of tests and be used  and this import does not really change something. Also is just a helper dataclass so we avoid reading data in tests from the tiledb.array for no reason when the only thing that matters and asserts is the schema of the data and not data per/se.

The distinction of unit and integration that I tried to impose here is that unit tests mock interactions with external dependencies (zarr, tiff etc..) whereas integration tests use the real dependencies which I think is pretty objective and not a matter of my opinion.

tests/integration/converters/__init__.py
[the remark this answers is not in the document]
gsakkis: There's no state and only a single method; ` CMU_1_SMALL_REGION().schemas()` is a verbose way to call a stateless `get_CMU_1_SMALL_REGION_schemas()`

tests/integration/converters/__init__.py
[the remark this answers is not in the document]
gsakkis: Because it removes the implicit assumption `len(_dom) == len(_tile)` and the hardcoded `for elem_id in range(0, 3)`. Adding (or removing) one schema now requires updating 3 lines instead of one.

tests/unit/test_openslide.py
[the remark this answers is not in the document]
gsakkis: > unit tests mock interactions with external dependencies

There's no mocking in this test file so according to this definition it should move to integrations.

tests/integration/converters/__init__.py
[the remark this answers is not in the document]
ktsitsi: Isn't there a 1-1 relation between domain and tile? Is it an assumption do we have a case? Anyway I will proceed to the change

tests/unit/converters/test_ome_zarr.py
gsakkis: They don't have to be realistic in the sense that they can be random-generated images but at least the array shape and dtype should be correct.

Also there's the issue of readability:
```
zarr_array = np.array(
    [[[[[[[0, 1], [2, 3]], [[4, 5], [6, 7, 8]]]]]]], dtype=object
)
...
expected = np.array(
    [[[[0, 1]], [[4, 5]]], [[[2, 3]], [[6, 7, 8]]]], dtype=object
)
```
I bet the `expected` value was produced by just running the test with the debugger and copying whatever result it returned, assuming it is correct. I strongly doubt a reviewer of test who just reads the input can come up with the expected result without running the code, it's not obvious at all.

tests/integration/converters/test_ome.py
```diff
@@ -25,32 +25,39 @@ def test_ome(format_path):
...
-    if not os.path.exists(dest):
-        OMETiffConverter().convert_image(src, dest, level_min=0)
-    with tiledb.open(os.path.join(dest, "l_0.tdb")) as A:
-        assert A.schema == expected[0]
-    with tiledb.open(os.path.join(dest, "l_1.tdb")) as A:
-        assert A.schema == expected[1]
-    with tiledb.open(os.path.join(dest, "l_2.tdb")) as A:
-        assert A.schema == expected[2]
+@pytest.mark.parametrize(
+    "converter",
```
gsakkis, line 42: [parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) accepts comma separated params, e.g. "converter,path" which avoids the need for `converter[0]/converter[1]`

tests/__init__.py
```diff
@@ -1,11 +1,85 @@
...
+            ],
+            cell_order="row-major",
+            tile_order="row-major",
+            capacity=10000,
+        )
+        for elem_id in range(len(_domains))
+    ]
+
+
+def _check_level_info(num, level_info):
```
gsakkis, line 76: s/_check_level_info/check_level_info; this is public now

tests/integration/converters/test_ome.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@pytest.mark.parametrize(
+    "converter,path",
+    [
+        (OMEZarrConverter, "CMU-1-Small-Region.ome.zarr"),
+        (OMETiffConverter, "CMU-1-Small-Region.ome.tiff"),
+    ],
+)
+def test_ome_tiff_converter(tmp_path, converter, path):
```
gsakkis, line 46: s/test_ome_tiff_converter/test_ome_converter

tests/unit/converters/test_openslide.py
```diff
@@ -0,0 +1,43 @@
...
+
+        # After substitution
+        reader._osd.level_count = 3
+        assert reader.level_count == 3
+
+    def test_osd_level_downsamples(self, mocker_osd):
+        reader = OpenSlideReader("")
+        reader._osd.level_downsamples = (5.6, 7.4, 6.7)
+        assert reader.level_downsamples == (5.6, 7.4, 6.7)
+        assert reader.level_downsamples != (5.6, 7.4, 6.8)
```
gsakkis, line 27: redundant assertion

gsakkis (approved):
Some trivial final comments, LGTM otherwise

PR #15, opened 2022-10-06
Change level_downsamples from stored attribute to computed property

gsakkis:
`level_downsamples` is [derived](https://github.com/openslide/openslide/blob/dd3a152250b460860614a3a793309245eebbcb7c/src/openslide.c#L285-L287) from `level_dimensions`, therefore it can be computed at runtime instead of being stored as metadata at ingestion time.

### DIFF
--- a/tiledbimg/openslide.py
+++ b/tiledbimg/openslide.py
@@ -10,9 +10,9 @@ import tiledb
 
 @dataclass(frozen=True)
 class LevelInfo:
-    uri: str = field(compare=False)
-    level: int = field(compare=True)
-    dimensions: Tuple[int, ...] = field(compare=True)
+    uri: str = field(compare=False, repr=False)
+    level: int
+    dimensions: Tuple[int, int]
 
     @classmethod
     def from_array(cls, array: tiledb.Array, level: Optional[int] = None) -> LevelInfo:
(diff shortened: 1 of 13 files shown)


ktsitsi:
I think this was there in accordance to the openslide API even though there is no real reason to keep it technically is it possible that we want to stay compatible? https://openslide.org/api/python/#openslide.OpenSlide.level_downsamples

tiledbimg/openslide.py
```diff
@@ -50,17 +48,27 @@ def level_count(self) -> int:
         return len(self.level_info)
 
     @property
-    def dimensions(self) -> Tuple[int, ...]:
+    def dimensions(self) -> Tuple[int, int]:
         """A (width, height) tuple for level 0 of the slide."""
-        return self.level_info[0].dimensions[:2]
+        return self.level_info[0].dimensions
```
ktsitsi, line 53: I assume the changes `dimensions[:2]` is to match the function description `(width, height)`  and in future PRs we will generalise again?

gsakkis: The [OpenSlide API](https://openslide.org/api/python/#openslide.OpenSlide.dimensions) handles a fixed number of dimensions (width, height) so we copy this over to `TileDBOpenSlide`. How to generalize this to more dimensions, or even if we should come up with a more generic API from scratch instead of trying to shoehorn it to OpenSlide is to be decided.

ktsitsi:
Maybe out of the scope of this ticket but since we increase test coverage. Should we import a `pytest.ini` to ignore coverage for e.g `tiledbimg/version.py`

PR #18, opened 2022-10-11
Support multi-channel images

gsakkis:
Add support for any number of channels instead of exactly 3 (RGB).

Note: format-breaking change


tiledbimg/converters/base.py
```diff
@@ -15,15 +16,31 @@ def level_count(self) -> int:
...
 
 
-class ImageConverter(ABC):
-    _rgb_dtype = np.dtype([("", "uint8"), ("", "uint8"), ("", "uint8")])
+@dataclass(frozen=True)
+class Dimension:
+    name: str
+    max_tile: int
+
+    def to_tiledb(self, size: int) -> tiledb.Dim:
```
georgeSkoumas, line 27: `to_tiledb_dim`?

georgeSkoumas (approved):

tiledbimg/converters/base.py
gsakkis: I thought It's implied since `self` is a dimension, can rename it if not obvious.

georgeSkoumas: For sure it's obvious, so your call. It's just an intuition of mine that `to_tiledb` = `to_tiledb_array`.

PR #19, opened 2022-10-12
Determine dtype dynamically

gsakkis:
Set the `dtype` of the TileDB dimensions and attribute based on the image numpy array instead of hardcoding them to uint64 and uint8, respectively. For dimensions specifically, the selected `dtype` is the smallest `dtype` that can hold `image.size` (`== np.prod(image.shape)`, the number of scalars in the array).


ktsitsi (approved):
Sweet

PR #20, opened 2022-10-19
Ingest OME-TIFF metadata

gsakkis:
- Add general support for ingesting group and array metadata to the (base) `ImageConverter`.
- Replace upstream `Tifffile` with a [fork](https://github.com/TileDB-Inc/tifffile/tree/gsa/python-3.7) that has been cloned from Tifffile `v2022.10.10` and backported to Python 3.7.
- Update `OMETiffReader` to: 
  - Read all TIFF pages, not just the top-level ones for each series. 
    **Note**: The image hierarchy is not currently preserved; the images are still stored as a flat list of TileDB arrays. 
  - Read the metadata needed to reconstruct (an approximation of) the original TIFF file and store them to TileDB group and array metadata as pickled bytestrings. The selected metadata are mostly based on the [`test_rewrite_ome` TIfffile test](https://github.com/cgohlke/tifffile/blob/a73812e98f58b068a590de862ed547820fd332ad/tests/test_tifffile.py#L16286-L16434).
 - Split converter tests to separate module per converter and delete the `tests/data/*tdb` TileDB arrays from the repo.


ktsitsi (approved):
LGTM!!

PR #25, opened 2022-10-20
Metadata tweaks

gsakkis:
Misc. metadata fixes and improvements based on real-world TIFF files:
- Ignore missing OME-XML metadata
- Drop the `original_filename` metadata key; YAGNI
- Handle properly axes transformations (swapaxes, moveaxis) for any 3D image
  - A more general and elegant solution to this problem will be addressed in a separate PR.
- Store image level in metadata (instead of trying to infer it from the uri).

### DIFF
--- a/tests/unit/test_openslide.py
+++ b/tests/unit/test_openslide.py
@@ -1,20 +1,25 @@
-import os
-
-import pytest
 import tiledb
 
-from tests import check_level_info, get_CMU_1_SMALL_REGION_schemas
-from tiledbimg.openslide import LevelInfo
+from tests import get_CMU_1_SMALL_REGION_schemas
+from tiledbimg.openslide import LevelInfo, TileDBOpenSlide
 
 
-class TestLevelInfo:
-    def test_from_array(self, tmp_path):
-        test_data_schemas = get_CMU_1_SMALL_REGION_schemas()
-        tiledb.Array.create(os.path.join(tmp_path, "test.tdb"), test_data_schemas[0])
-        with tiledb.open(os.path.join(tmp_path, "test.tdb"), "r") as A:
-            for level in range(42):
-                check_level_info(level, LevelInfo.from_array(A, level))
+class TestTileDBOpenSlide:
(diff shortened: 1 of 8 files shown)


tests/unit/test_openslide.py
```diff
@@ -1,20 +1,25 @@
...
+    def test_from_group_uri(self, tmp_path):
+        schemas = get_CMU_1_SMALL_REGION_schemas()
+        group_path = str(tmp_path)
+        tiledb.Group.create(group_path)
+        with tiledb.Group(group_path, "w") as G:
+            for level, schema in enumerate(schemas):
+                level_path = str(tmp_path / f"l_{level}.tdb")
+                tiledb.Array.create(level_path, schema)
+                with tiledb.open(level_path, "w") as A:
+                    A.meta["level"] = level
```
Shelnutt2, line 17: Should we also use the group name to add the member to the group with a level?

gsakkis: You mean `G.add(level_path)`? In principle we should (and `ImageConverter.convert_image()` does it) but the test is passing even without it. So It looks as if every array under the group directory is implicitly a group member? In any case I can add it explicitly for clarity.

Shelnutt2: I mean the `add` function for groups has three parameters. `uri`, `name` and `relative`. `name` lets you request the group member by name.
https://github.com/TileDB-Inc/TileDB-Py/blob/f8cca3167ea1bb3067fc1bf60607979653b0b75f/tiledb/group.py#L291

```
with tiledb.Group(grp_uri, mode="w") as g:
   g.add(level_1, "level_1", relative=False)
...
with tiledb.Group(grp_uri, mode="r") as g:
   level_1_uri = g["level_1"]
```

Shelnutt2 (approved):

tests/unit/test_openslide.py
gsakkis: I didn't know about the `name` parameter, that's good to know. For now it's not needed but will keep it in mind as it may simplify things internally.

PR #26, opened 2022-10-31
Change dimension order from XYC to CYX

gsakkis:
First PR towards supporting any NGFF compatible 2 to 5 dimensional image data; more to follow.

This PR still supports only 3D images but changes the order of dimensions from the current `XYC` (width, height, channel) to `CYX`. This is the canonical dimension order of multichannel images (the full canonical order for 5D data is `TCZYX`).

Additionally, `TileDBOpenSlide.read_region` changed to return a 3D numpy array in `YXC` order instead of `XYC` in order to be consistent with the Numpy ([as opposed to PIL](https://stackoverflow.com/questions/57259519/handling-height-width-vs-width-height-coordinate-order-in-pil-vs-ndarray)) convention.

**Note**: backwards incompatible change.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -21,8 +21,7 @@ def rgb_to_5d(pixels: np.ndarray) -> Sequence:
         channels = np.array([stack])
     elif len(pixels.shape) == 3:
         size_c = pixels.shape[2]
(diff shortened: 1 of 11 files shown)


ktsitsi (approved):
LGTM

PR #27, opened 2022-11-02
Support general axes transpositions

gsakkis:
Second PR towards supporting any NGFF compatible 2 to 5 dimensional image data; follow-up to https://github.com/TileDB-Inc/TileDB-BioImaging/pull/26.

This PR:
- Moves the swap/move axes logic (currently hardcoded in the various converters) to a new `tiledbimg.converters.axes` module
- Generalizes this logic to support transforming from an arbitrary axes sequence to a canonical axes sequence. Ideally the transformation uses the minimal number of transpositions, although this is unproven; the current algorithm just picks greedily the transposition that minimizes the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) between the input and the target sequence.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,6 +13,6 @@ repos:
     hooks:
     - id: flake8
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.961
+    rev: v0.982
     hooks:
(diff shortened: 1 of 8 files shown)


ktsitsi (approved):
LGTM

PR #29, opened 2022-11-08
Misc. converter and test updates

gsakkis:
- Update test data to have a common source:
  - `CMU-1-Small-Region.svs`: Downloaded from https://openslide.cs.cmu.edu/download/openslide-testdata/Aperio/CMU-1-Small-Region.svs and used by `OpenSlideConverter`
  - `CMU-1-Small-Region.ome.zarr`: Derived from `CMU-1-Small-Region.svs` by the [bioformats2raw](https://github.com/glencoesoftware/bioformats2raw) converter and used by `OMEZarrConverter`
  - `CMU-1-Small-Region.ome.tiff`: Derived from `CMU-1-Small-Region.ome.zarr` by the [raw2ometiff](https://github.com/glencoesoftware/raw2ometiff) converter and used by `OMETiffConverter`
- Update `OMETiffConverter` to read only the pyramid levels of the first series instead of making a flat list of all levels of all series.
- Fix `OMEZarrConverter` to preserve more metadata and read levels in order
- Refactor OMEZarrReader
- Refactor `TileDBOpenSlide`: drop `level_info` attribute and `LevelInfo` dataclass
- Misc. test cleanups


tiledbimg/converters/ome_zarr.py
```diff
@@ -115,12 +119,18 @@ def level_metadata(self, level: int) -> Dict[str, Any]:
         return {"pickled_zarrwriter_kwargs": pickle.dumps(writer_kwargs)}
 
     def metadata(self) -> Dict[str, Any]:
-        multiscale_meta = self.image.zarr.root_attrs.get("multiscales")
-        assert len(multiscale_meta) == 1
+        multiscales = self.image.zarr.root_attrs.get("multiscales")
+        assert len(multiscales) == 1, multiscales
+        multiscale = multiscales[0]
+        coordinate_transformations = (
```
ktsitsi, line 125: Why not with `get_coord = operator.itemgetter("coordinateTransformations")` and `list(map(get_coord, multiscale.get("datasets")))`

tests/unit/test_openslide.py
```diff
@@ -1,12 +1,18 @@
...
+from tiledbimg.openslide import TileDBOpenSlide
 
 
 class TestTileDBOpenSlide:
     def test_from_group_uri(self, tmp_path):
-        schemas = get_CMU_1_SMALL_REGION_schemas()
+        def r():
+            return random.randint(64, 4096)
+
+        level_dimensions = [(r(), r()) for _ in range(7)]
```
ktsitsi, line 14: Can you add a comment here on why hard coding `range(7)`

gsakkis: No particular reason, I can make this random too

tiledbimg/converters/ome_zarr.py
gsakkis: To prevent KeyError in case `coordinateTransformations` is missing from a dataset dict

tests/unit/test_openslide.py
gsakkis: [Done](https://github.com/TileDB-Inc/TileDB-BioImaging/pull/29/commits/3832f82dc7ce59d0b1a4dcddaafd50edde45aaf1)

tiledbimg/converters/ome_zarr.py
ktsitsi: According to NGFF spec here, I think we should error out in this cases but I leave it to your judgement.

> Each dictionary in "datasets" MUST contain the field "coordinateTransformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.

gsakkis: Let's ignore such dataset dicts for now since [ome-zarr-py](https://github.com/ome/ome-zarr-py/blob/master/ome_zarr/writer.py#L142) does the same.

ktsitsi (approved):

PR #31, opened 2022-11-09
Refactor ImageWriter/ImageConverter.from_tiledb

gsakkis:
- Refactor `ImageWriter` and `ImageConverter.from_tiledb` (and at a lesser degree `ImageReader` and `ImageConverter.to_tiledb`).
- Change the OME-Zarr stored metadata from pickle to json.


ktsitsi (approved):
👍

PR #32, opened 2022-11-11
BioImaging API DOCS

ktsitsi:
This PR:

- Creates a `docs` directory to hold all `API` documentation
- Modifies the functions' docstrings of most of the modules under `tiledbimg` package
- Imports static logos from TILEDB registry
- Auto generates  files by ` sphinx-apidoc` 
- Modifies `README.md` with instructions on building the docs


tiledbimg/converters/ome_tiff.py
```diff
@@ -20,15 +26,41 @@ def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
 
     @property
     def level_count(self) -> int:
+        """
```
gsakkis, line 29: Let's not copy-paste the docstrings from base classes, that's not sustainable.

tiledbimg/converters/base.py
```diff
@@ -150,6 +155,12 @@ def to_tiledb(
     def _write_image(
         self, uri: str, image: np.ndarray, metadata: Dict[str, Any]
     ) -> None:
+        """
```
gsakkis, line 158: No need to document (every) private methods (even more so when there's no actual docstring)

gsakkis (approved):

PR #38, opened 2022-11-21
General 2D-5D axes support

gsakkis:
Third and final PR towards supporting any NGFF compatible 2 to 5 dimensional image data; follow-up to https://github.com/TileDB-Inc/TileDB-BioImaging/pull/27.

- Determine the TileDB schema dynamically based on the axes, either the original (`preserve_axes=True`) or the canonical (`preserve_axes=False`). 
- Remove 3D assumption/restriction from converters and `TileDBOpenSlide`; support any valid 2D-5D axes
- Store the original axes in metadata and update the `ImageWriter`s to write the image back to its original axes.
- Various internal updates, refactorings and tests to support the above.


tests/unit/test_axes.py
```diff
@@ -64,104 +64,202 @@ def test_init(self):
...
+            assert_canonical_transpose(s, a, np.swapaxes(np.moveaxis(a, 3, 1), 2, 3))
         for s in "ZXYC", "CXYT", "ZXYT":
-            assert_transpose(s, a, np.swapaxes(np.moveaxis(a, 3, 0), 2, 3))
+            assert_canonical_transpose(s, a, np.swapaxes(np.moveaxis(a, 3, 0), 2, 3))
         for s in "XYCZ", "XYTC", "XYTZ":
-            assert_transpose(s, a, np.swapaxes(np.moveaxis(a, 2, 0), 1, 3))
+            assert_canonical_transpose(s, a, np.swapaxes(np.moveaxis(a, 2, 0), 1, 3))
 
-    def test_transpose_5d(self):
+    def test_transpose_canonical_5d(self):
```
ktsitsi, line 155: Could we keep one naming convention like `test_canonical_transpose_5d`?

tiledb/bioimg/converters/axes.py
```diff
@@ -49,6 +49,32 @@ def transposed_array(self, a: np.ndarray) -> np.ndarray:
         return np.moveaxis(a, self.i, self.j)
 
 
+def transpose_array(a: np.ndarray, s: Sequence[T], t: Sequence[T]) -> np.ndarray:
```
ktsitsi, line 52: Does this need to be accessible outside the `Axes` dataclass ?

tiledb/bioimg/converters/base.py
```diff
@@ -37,49 +22,34 @@ def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
...
+    def axes(self) -> Axes:
+        """The axes of this multi-resolution image."""
 
+    @property
     @abstractmethod
-    def level_axes(self, level: int) -> Axes:
-        """Return the axes for the given level.
-
-        :param level: Number corresponding to a level
+    def level_count(self) -> int:
```
ktsitsi, line 30: Have you tried building the html page with the updated docstrings? All the derived classes get the `docstrings` from `base` class `docstrings`.  `Docstrings` follow `reStructuredText` format do we want to change that?

tiledb/bioimg/converters/base.py
```diff
@@ -157,39 +130,62 @@ def to_tiledb(
...
-                level_metadata = reader.level_metadata(level)
-                level_metadata["level"] = level
-                self._write_image(uri, canonical_image, level_metadata)
+                # determine axes and (optionally) transpose image to canonical axes
+                if preserve_axes:
+                    level_axes = axes
+                else:
+                    level_axes = axes.canonical(image)
+                    image = axes.transpose(image, level_axes)
+                # create TileDB array
```
ktsitsi, line 152: Why unwrapping the `_write_image`?

tiledb/bioimg/converters/base.py
gsakkis: Just tried it, looks good to me. It can still be parsed as reStructuredText, I just removed the duplicated `:param:`/`:return:` text that's already in the method description. 
Screenshot from 2022-11-21 20-07-26

tiledb/bioimg/converters/axes.py
gsakkis: Dropped `Axes.transpose` at https://github.com/TileDB-Inc/TileDB-BioImaging/pull/38/commits/5e5c5734fcbdd26ea905e3d6289e9b4a4deb06ea

tiledb/bioimg/converters/base.py
ktsitsi: It would render for sure. I get that these are `small` methods or properties and that `params` and `returns` feel verbose, but it creates this heterogeneity between `write_level_image` and `write_group_metadata` for example where in the `metadata`  receive explanation in the former but not in the latter. And secondly imo each API should be self-contained and self-explanatory without the need to look somewhere else in the docs. Anyway your call at this dev stage probably it does not make any difference.

tiledb/bioimg/converters/base.py
ktsitsi: Yeap I was thinking mostly about your second point. Define/Get the `schema` outside and pass it to `_write_image` were the logic of writing will be held.

tiledb/bioimg/converters/base.py
gsakkis: Ok I see what you mean, indeed this trades off the redundancy in docstrings with occasional inconsistency; some parameters are self-explanatory, some aren't (though in the example you mentioned i'd rather remove the verbosity  in `write_level_image` than bring it back for all the rest).

As for the second point, what is not self-contained/self-explanatory? There's no linking to docs elsewhere afaict.

ktsitsi: I mean easy indexing as a user if i was about to look up at a function signature and only that I would like to see all arguments explained in the docstrings of the function without the need to look up somewhere else to find out what does metadata means, or which metadata we refer to for example. e.g Imagine `numpy` [transpose function](https://numpy.org/doc/stable/reference/generated/numpy.transpose.html) everybody knows what `axes` knows and it is used in many functions around the package. However as I said at this stage of dev might not make any sense and this might also be my personal preference so your call :)

gsakkis: Sure, no disagreement about that. I'm only arguing against the redundancy within the same docstring that doesn't add any new information:
```
def f(x,y):
  """
  Returns  {blahblah} given {the Foo x} and {the bar y}.

 :param x: {the Foo x}
 :param y: {the Bar y}
  
  :return: {blahblah}
  """
````

PR #39, opened 2022-11-21
Drop ImageConverter

gsakkis:
`ImageConverter` (& subclasses) doesn't really carry its weight:
- its instances have minimal state (zero after https://github.com/TileDB-Inc/TileDB-BioImaging/pull/38) 
- their two methods `to_tiledb` and `from_tiledb` are pretty much independent and better exposed as classmethods to `ImageReader` and `ImageWriter` respectively.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,23 +49,20 @@ How to convert imaging data from standard biomedical formats to group of TileDB 
 
 ### OME-Zarr to TileDB Group of Arrays
 ```python
-from tiledb.bioimg.converters.ome_zarr import OMEZarrConverter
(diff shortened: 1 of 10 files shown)


ktsitsi (commented):
LGTM but a general question for discussion... isn't it ambiguous what we mean by `ImageWriter` and `ImageReader`? Previously under `ImageConverter` someone intuitively would expect to find 2 API calls `from` and `to`, but now for someone external `ImageWriter` and `ImageReader` means pretty much nothing. Reading from where and writing from where? He/she should check the methods inside the classes to grasp the semantics.

PR #42, opened 2022-11-24
Add test for OMETiffConverter.to_tiledb() supporting fileobj source

gsakkis:


gsakkis:
Rebased with https://github.com/TileDB-Inc/TileDB-BioImaging/pull/43

PR #44, opened 2022-11-29
Replace Levenshtein dependency

gsakkis:
Replace [Levenshtein](https://pypi.org/project/Levenshtein/) (GPL) with ~[edit-operation](https://pypi.org/project/edit-operation/)~ [pyeditdistance](https://pypi.org/project/pyeditdistance/) (MIT). 

Note: Unlike `Levenshtein` that can compare arbitrary sequences, `edit-operation` can compare only strings, so the `Transpose` API and related functions were adjusted accordingly.

### DIFF
--- a/tiledb/bioimg/converters/axes.py
+++ b/tiledb/bioimg/converters/axes.py
@@ -6,8 +6,8 @@ from dataclasses import dataclass
 from operator import itemgetter
 from typing import Iterable, Iterator, List, MutableSequence, Sequence, TypeVar
 
-import Levenshtein
 import numpy as np
+from edit_operation import levenshtein
 
 T = TypeVar("T")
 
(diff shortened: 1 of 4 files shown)


tiledb/bioimg/converters/axes.py
```diff
@@ -87,19 +85,19 @@ def transpose_array(a: np.ndarray, s: Sequence[T], t: Sequence[T]) -> np.ndarray
...
-    t = list(t)
+    sbuf = bytearray(s.encode())
+    tbuf = t.encode()
     transpositions = []
-    while s != t:
+    while sbuf != tbuf:
         weighted_transpositions = (
-            (Levenshtein.distance(transposition.transposed(s), t), transposition)
-            for transposition in gen_transpositions(n)
+            (levenshtein.distance(tr.transposed(sbuf).decode(), t), tr)
```
georgeSkoumas, line 96: Any gain in case we parallelize here?

georgeSkoumas (approved):

tiledb/bioimg/converters/axes.py
gsakkis: Haven't checked but it's not worth it, the compared strings are tiny (2-5 chars long) and `levenshtein.distance` is in C++/Cython so the whole `minimize_transpositions` call takes less than a msec.

PR #46, opened 2022-11-29
Replace openslide-python with tiffslide

gsakkis:
Swap [openslide-python](https://github.com/openslide/openslide-python) with [tiffslide](https://github.com/bayer-science-for-a-better-life/tiffslide), a drop-in replacement based on [tifffile](https://github.com/cgohlke/tifffile) instead of the [OpenSlide](https://openslide.org/) C library.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 import setuptools
 
 zarr = ["ome-zarr"]
-openslide = ["openslide-python"]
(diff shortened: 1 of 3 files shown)


gsakkis:
Deferring for now since TiffSlide has not implemented support for all formats that openslide supports.

PR #48, opened 2022-12-07
Chunked ingestion

gsakkis:
Add initial support for ingestion of images in chunks to limit memory requirements.

The current API is minimal: an optional boolean `chunked` parameter in `ImageConverter.to_tiledb` that when true, each image level is read in chunks, where each chunk is a single tile region in TileDB. Later this may be extended to allow chunks larger than a single tile.

To enable chunked reading, the base `ImageReader` and all its subclasses have been extended:
- `level_image` now takes an optional `tile` parameter for reading a subregion of the image
- new `level_shape` method to return the shape of the image at the given level
- new `level_dtype` method to return the `dtype` of the image at the given level

Implementation notes:
- This PR allows chunked ingestion **only** if the original axes are preserved (either due to `preserve_axes=True` or because the original axes happens to be already in canonical order). A subsequent PR will lift this constraint. 
- The chunked read support for OME-Tiff requires `zarr` since [tifffile doesn't support random access directly](https://github.com/cgohlke/tifffile/issues/134).

### DIFF
--- a/tests/integration/converters/test_ome_tiff.py
+++ b/tests/integration/converters/test_ome_tiff.py
@@ -55,13 +55,21 @@ def test_ome_tiff_converter_different_dtypes(tmp_path):
         assert A.attr(0).dtype == np.uint16
 
 
-def test_tiledb_to_ome_tiff_rountrip(tmp_path):
+@pytest.mark.parametrize("preserve_axes", [False, True])
+@pytest.mark.parametrize("chunked", [False, True])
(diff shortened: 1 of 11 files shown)


georgeSkoumas (approved):
LGTM! 

Maybe the whole functionality in `tiles.py`,  could be removed (with the corresponding unit tests), because they are mainly one liners. Your call ;).

gsakkis:
Merged into #50

PR #50, opened 2022-12-13
Chunked ingestion with axes transpositions

gsakkis:
Follow-up to https://github.com/TileDB-Inc/TileDB-BioImaging/pull/48 to support chunked ingestion when the original axes are different from the saved (canonical) axes.


georgeSkoumas (approved):

ihnorton (commented):
Test question: if we are not preserving the axes, is there an implicit reordering of the zarr data which makes [these comparisons](https://github.com/TileDB-Inc/TileDB-BioImaging/pull/50/files#diff-8381c7f7e0dbdb7df612403cde77c753fe2e93b5f1d9f54e8405201914d35e07R83-R88) succeed?

PR #54, opened 2022-12-21
Support multithreaded chunked ingestion

gsakkis:
- Support parallel ingestion on multiple threads by passing a `max_workers` parameter.
- Display a progress bar during conversion (via [tqdm](https://github.com/tqdm/tqdm)): progressbar

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ setuptools.setup(
         "local_scheme": "dirty-tag",
         "write_to": "tiledb/bioimg/version.py",
     },
-    install_requires=["pyeditdistance", "tiledb"],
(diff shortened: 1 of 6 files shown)


ktsitsi:
Could we have a verification test between single and multithreaded `to_tiledb` conversion just for completeness? Otherwise LGTM!

gsakkis:
> Could we have a verification test between single and multithreaded `to_tiledb` conversion just for completeness? Otherwise LGTM!

Although there is no explicit comparison between single and multithreaded conversion in the tests, the parametric tests actually offer a stronger guarantee: that in all cases the conversion is correct (at least as far as each test goes to verify correctness).

ktsitsi (approved):

PR #56, opened 2023-01-05
Inserting bioimg version in group metadata

ktsitsi:
This PR:

- Add an extra group metadata key for tagging version of `tiledb.bioimg`. Now the group metadata of the converted images are ```{'axes': 'YXC', 'version': '0.1.1'}```

### DIFF
--- a/tiledb/bioimg/converters/base.py
+++ b/tiledb/bioimg/converters/base.py
@@ -23,6 +23,8 @@ import tiledb
 from .axes import Axes, AxesMapper
 from .tiles import iter_tiles, num_tiles
 
+FMT_VERSION = 1
+
 
 class ImageReader(ABC):
     @abstractmethod
@@ -251,7 +253,8 @@ class ImageConverter:
                 group.meta.update(
                     reader.group_metadata,
                     axes=input_axes.dims,
-                    version=".".join(map(str, bioimg_version[0:3])),
+                    pckg_version=".".join(map(str, bioimg_version[0:3])),
+                    fmt_version=FMT_VERSION,
                 )
                 for uri in uris:
                     if urlparse(uri).scheme == "tiledb":
--- a/tiledb/bioimg/converters/base.py
+++ b/tiledb/bioimg/converters/base.py
@@ -11,6 +11,8 @@ from urllib.parse import urlparse
 import numpy as np
 from tqdm import tqdm
 
+from ..version import version_tuple as bioimg_version
+
 try:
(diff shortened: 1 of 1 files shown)


tiledb/bioimg/converters/base.py
```diff
@@ -247,6 +249,7 @@ def tile_to_tiledb(level_tile: Tuple[slice, ...]) -> None:
             # Write group metadata
             with tiledb.Group(output_path, "w") as group:
                 group.meta.update(reader.group_metadata, axes=input_axes.dims)
+                group.meta.update({"version": ".".join(map(str, bioimg_version[0:3]))})
```
gsakkis, line 252: This can be moved to the previous update call (`update(axes=..., version=...)`)

tiledb/bioimg/converters/base.py
```diff
@@ -246,7 +248,11 @@ def tile_to_tiledb(level_tile: Tuple[slice, ...]) -> None:
 
             # Write group metadata
             with tiledb.Group(output_path, "w") as group:
-                group.meta.update(reader.group_metadata, axes=input_axes.dims)
+                group.meta.update(
+                    reader.group_metadata,
+                    axes=input_axes.dims,
+                    version=".".join(map(str, bioimg_version[0:3])),
```
gsakkis, line 254: Why not just `__version__`?

gsakkis (approved):

tiledb/bioimg/converters/base.py
ktsitsi: Was just thinking/overthinking that we might want to eliminate `dirty` versions. Another thing is that even in that form patch version cannot dictate a specification change. I would assume spec would change only in minor-or even major version.

tiledb/bioimg/converters/base.py
```diff
@@ -246,7 +250,12 @@ def tile_to_tiledb(level_tile: Tuple[slice, ...]) -> None:
 
             # Write group metadata
             with tiledb.Group(output_path, "w") as group:
-                group.meta.update(reader.group_metadata, axes=input_axes.dims)
+                group.meta.update(
+                    reader.group_metadata,
+                    axes=input_axes.dims,
+                    pckg_version=version,
```
gsakkis, line 256: Is this deliberately spelled `pckg` or a typo for `pkg`?

ktsitsi: It was deliberately but let's use a 3-letter abbreviation.

PR #57, opened 2023-01-12
Add TileDBOpenSlide.read_level

gsakkis:
Add a `TileDBOpenSlide.read_level` method for the (common) case of selecting a whole level image rather than a specific region with `read_region`.


ktsitsi (approved):
🚀

PR #60, opened 2023-01-13
Xan/sc 24385/tile server

XanthosXanthopoulos:
Currently we have not any proper data to test the server metadata and slice methods. I set this as a draft since there are no test implemented

### DIFF
--- a/tiledb/bioimg/utils/tiles_server.py
+++ b/tiledb/bioimg/utils/tiles_server.py
@@ -1,6 +1,6 @@
 import io
+import platform
 
-# import os
 # import glob
 from dataclasses import dataclass
 
@@ -8,11 +8,13 @@ from dataclasses import dataclass
 from typing import Any, Callable, Dict, List, Tuple
 
 import cv2
+import flask
 import numpy as np
-from flask import Flask, Response, jsonify, make_response  # , send_file
+from flask import Flask, Response, jsonify, make_response
 from flask_cors import CORS
 from multiprocess import Pool
 from PIL import Image
+from werkzeug.serving import run_simple
 
 import tiledb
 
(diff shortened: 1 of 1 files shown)


tiledb/bioimg/utils/tiles_server.py
```diff
@@ -0,0 +1,276 @@
...
+            for record in self._metadata
+        }
+
+        CORS(self._app)
+
+    def __del__(self) -> None:
+        self._pool.terminate()
+
+    def run(
+        self, host: str = "0.0.0.0", port: int = 8000, multiprocess: bool = False
```
gsakkis, line 222: Generalize `multiprocess: bool = False` to `num_processes: int = 0`

tiledb/bioimg/utils/tiles_server.py
```diff
@@ -0,0 +1,276 @@
...
+    ) -> None:
+        print(platform.system())
```
gsakkis, line 224: leftover print

tiledb/bioimg/utils/tiles_server.py
```diff
@@ -0,0 +1,276 @@
...
+        if platform.system() == "Windows":
+            self._pool.apply_async(self._app.run, kwds={"host": host, "port": port})
```
gsakkis, line 226: What's wrong with `run_simple` on Windows?

gsakkis (changes requested):
Looks great overall :100:  I left a few code style comments but nothing major.

Two more general comments:
- This doesn't have to be a package (also `utils` is usually a name for internal helpers). A plain `tiledb/bioimg/tiles_server.py` script will do.
- It should be packaged as an optional dependency and its dependencies (flask etc) be added to `extras_require` in setup.py.

tiledb/bioimg/utils/tiles_server.py
```diff
@@ -0,0 +1,276 @@
...
+
+# from pathlib import Path
+from typing import Any, Callable, Dict, List, Tuple
+
+import cv2
+import flask
+import numpy as np
+from flask import Flask, Response, jsonify, make_response
+from flask_cors import CORS
+from multiprocess import Pool
```
gsakkis, line 15: `s/multiprocess/multiprocessing`

XanthosXanthopoulos:
The tile server will not be used for now so the PR closed

PR #61, opened 2023-01-24
Xan/sc 23635/pyramid generation by downsampling input

XanthosXanthopoulos:
There are four different ways to generate downsampled version of the original image. Each method however generates different images if we compare the raw pixel values but there is no ground truth to evaluate which is better. 

From the four methods the `CHUNKED_PROGRESSIVE` is the most memory efficient and fast. All other methods are prone to crashing if the is not enough memory. We currently are not testing images with more than 3 dimensions.

### DIFF
--- a/tests/integration/converters/test_openslide.py
+++ b/tests/integration/converters/test_openslide.py
@@ -1,9 +1,9 @@
+import numpy as np
 # import os
 #
 # os.add_dll_directory("C:/openslide/bin")
 import json
 
-import numpy as np
 import openslide
 import PIL.Image
 import pytest
--- a/tests/integration/converters/test_openslide.py
+++ b/tests/integration/converters/test_openslide.py
@@ -1,9 +1,6 @@
-import numpy as np
-# import os
-#
-# os.add_dll_directory("C:/openslide/bin")
 import json
 
+import numpy as np
 import openslide
 import PIL.Image
 import pytest
--- a/tests/integration/converters/test_openslide.py
+++ b/tests/integration/converters/test_openslide.py
@@ -1,3 +1,6 @@
+# import os
+#
+# os.add_dll_directory("C:/openslide/bin")
 import json
 
 import numpy as np
--- a/tiledb/bioimg/converters/base.py
+++ b/tiledb/bioimg/converters/base.py
@@ -6,7 +6,7 @@ from abc import ABC, abstractmethod
 from concurrent.futures import ThreadPoolExecutor
 from typing import Any, Dict, Mapping, Optional, Tuple, Type
 from operator import itemgetter
-from typing import Any, Dict, Mapping, Optional, Tuple, Type, List
+from typing import Any, Dict, List, Mapping, Optional, Tuple, Type
 from urllib.parse import urlparse
 
 import numpy as np
@@ -14,7 +14,7 @@ from tqdm import tqdm
 
 from ..version import version
 
-from .scale import Scaler, ScaleMethod, ScalerMode
+from .scale import ScaleMethod, Scaler, ScalerMode
 
 try:
     from tiledb.cloud import groups
@@ -162,7 +162,7 @@ class ImageConverter:
         compressor: tiledb.Filter = tiledb.ZstdFilter(level=0),
         generate_pyramid: bool = False,
         pyramid_mode: ScalerMode = ScalerMode.CHUNKED_PROGRESSIVE,
-        pyramid_scale: List[int] = None,
+        pyramid_scale: List[int] = [],
         interpolation_method: ScaleMethod = ScaleMethod.NEAREST,
         register_kwargs: Optional[Mapping[str, str]] = {},
     ) -> None:
@@ -211,8 +211,10 @@ class ImageConverter:
             uris = []
             level_max = reader.level_count if not generate_pyramid else level_min + 1
             if reader.level_count > 1 and generate_pyramid:
-                print("Warning: The image contains multiple levels but pyramid generation is enabled. All levels "
-                      "except level zero will be skipped")
+                print(
+                    "Warning: The image contains multiple levels but pyramid generation is enabled. All levels "
+                    "except level zero will be skipped"
+                )
 
             for level in range(level_min, level_max):
             levels_meta = []
@@ -289,7 +291,11 @@ class ImageConverter:
                         a[:] = axes_mapper.map_array(image)
 
                     if generate_pyramid:
-                        scaler = Scaler(a, Axes("".join(dim.name for dim in a.domain)), scale_factor=pyramid_scale)
+                        scaler = Scaler(
+                            a,
+                            Axes("".join(dim.name for dim in a.domain)),
+                            scale_factor=pyramid_scale,
+                        )
                 uris.append(uri)
 
             if generate_pyramid:
@@ -304,14 +310,27 @@ class ImageConverter:
                         max_tiles=ChainMap(dict(tiles or {}), cls._DEFAULT_TILES),
                     )
                     tiledb.Array.create(uri, schema)
-                    if pyramid_mode == ScalerMode.NON_PROGRESSIVE or pyramid_mode == ScalerMode.CHUNKED_NON_PROGRESSIVE:
+                    if (
+                        pyramid_mode == ScalerMode.NON_PROGRESSIVE
+                        or pyramid_mode == ScalerMode.CHUNKED_NON_PROGRESSIVE
+                    ):
                         with tiledb.open(uris[0], "r") as r:
                             with tiledb.open(uri, "w") as a:
                                 a.meta.update(metadata, level=level)
                                 if pyramid_mode == ScalerMode.NON_PROGRESSIVE:
-                                    scaler.apply(r, a, level - level_min - 1, interpolation_method)
+                                    scaler.apply(
+                                        r,
+                                        a,
+                                        level - level_min - 1,
+                                        interpolation_method,
+                                    )
                                 else:
-                                    scaler.apply_chunked(r, a, level - level_min - 1, interpolation_method)
+                                    scaler.apply_chunked(
+                                        r,
+                                        a,
+                                        level - level_min - 1,
+                                        interpolation_method,
+                                    )
                     else:
                         with tiledb.open(uris[-1], "r") as r:
                             with tiledb.open(uri, "w") as a:
@@ -319,7 +338,9 @@ class ImageConverter:
                                 if pyramid_mode == ScalerMode.PROGRESSIVE:
                                     scaler.apply_progressive(r, a, interpolation_method)
                                 else:
-                                    scaler.apply_chunked_progressive(r, a, interpolation_method)
+                                    scaler.apply_chunked_progressive(
+                                        r, a, interpolation_method
+                                    )
                     uris.append(uri)
                     level += 1
 
--- a/tiledb/bioimg/converters/base.py
+++ b/tiledb/bioimg/converters/base.py
@@ -5,10 +5,16 @@ import os
 from abc import ABC, abstractmethod
 from concurrent.futures import ThreadPoolExecutor
 from typing import Any, Dict, Mapping, Optional, Tuple, Type
+from operator import itemgetter
+from typing import Any, Dict, Mapping, Optional, Tuple, Type, List
 from urllib.parse import urlparse
 
 import numpy as np
 from tqdm import tqdm
+
+from ..version import version
+
+from .scale import Scaler, ScaleMethod
 
 try:
     from tiledb.cloud import groups
@@ -154,6 +160,9 @@ class ImageConverter:
         register_kwargs: Mapping[str, Any] = {},
         reader_kwargs: Mapping[str, Any] = {},
         compressor: tiledb.Filter = tiledb.ZstdFilter(level=0),
+        generate_pyramid: bool = False,
+        pyramid_scale: float | List[float] = 1.0,
+        register_kwargs: Optional[Mapping[str, str]] = {},
     ) -> None:
         """
         Convert an image to a TileDB Group of Arrays, one per level.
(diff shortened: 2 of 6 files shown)


tiledb/bioimg/converters/base.py
```diff
@@ -252,6 +270,60 @@ def tile_to_tiledb(level_tile: Tuple[slice, ...]) -> None:
...
+                            a,
+                            Axes("".join(dim.name for dim in a.domain)),
+                            scale_factor=pyramid_scale,
+                        )
+                uris.append(uri)
+
+            if generate_pyramid:
+                level = level_min + 1
+
+                for dimension in scaler.resolutions:
```
gsakkis, line 284: The body of this loop should be:
(a) extracted to a private method; this method is already long/complex enough without all the pyramid generation logic
(b) refactored to eliminate all duplicated logic (e.g. between the `NON_PROGRESSIVE` vs `CHUNKED_NON_PROGRESSIVE` )

tiledb/bioimg/converters/scale.py
```diff
@@ -0,0 +1,272 @@
...
+            ratio = scale_factor[level]
+
+            shape = list(self._initial_shape)
+
+            shape[self._x_index] = int(
+                round(self._initial_shape[self._x_index] / float(ratio))
+            )
+            shape[self._y_index] = int(
+                round(self._initial_shape[self._y_index] / float(ratio))
+            )
```
gsakkis, line 57: - No need for `float()`
- No need for `int()`: `round()` already returns an int (in Python 3)

tiledb/bioimg/converters/scale.py
```diff
@@ -0,0 +1,272 @@
...
+class Scaler(object):
+    def __init__(
+        self, base: tiledb.array.DenseArray, axes: Axes, scale_factor: List[int]
+    ):
+        self._scale_factor = scale_factor
+        self._scale_factor_progressive = [float(scale_factor[0])]
+
+        for index in range(1, len(scale_factor)):
+            self._scale_factor_progressive.append(
+                scale_factor[index] / float(scale_factor[index - 1])
```
gsakkis, line 36: No need for `float()`: division is not truncating (in Python 3)

tiledb/bioimg/converters/scale.py
```diff
@@ -0,0 +1,272 @@
...
+class ScalerMode(Enum):
+    NON_PROGRESSIVE = 0
+    PROGRESSIVE = 1
+    CHUNKED_NON_PROGRESSIVE = 2
+    CHUNKED_PROGRESSIVE = 3
+
+
+class Scaler(object):
+    def __init__(
+        self, base: tiledb.array.DenseArray, axes: Axes, scale_factor: List[int]
```
gsakkis, line 29: - `tiledb.Array` for annotating TileDB arrays, don't care if it's dense or sparse (same comment for all `tiledb.array.DenseArray` occurrences)
- `Sequence` instead of `List` for immutable sequences (same comment for all `List` occurrences)

tiledb/bioimg/converters/scale.py
```diff
@@ -0,0 +1,272 @@
...
+    ) -> None:
+        if input_data.shape != self._initial_shape:
+            raise ValueError(
+                f"Incompatible input shape. Expected {self._initial_shape}, got {input_data.shape}"
+            )
+
+        if level >= len(self._resolutions):
+            raise ValueError(
+                f"Unknown zoom level. Expected 0 to {len(self._resolutions) - 1}, got {level}"
+            )
```
gsakkis, line 100: Strictly speaking not necessary, you'll get an IndexError in `self._resolutions[level]`.

tiledb/bioimg/converters/scale.py
```diff
@@ -0,0 +1,272 @@
...
+                    scale_method=method.value,
+                ),
+                iter_tiles(out_data.domain),
+            )
+
+    def apply_chunked_progressive(
+        self,
+        base: tiledb.array.DenseArray,
+        out_data: tiledb.array.DenseArray,
+        method: ScaleMethod,
```
gsakkis, line 169: Please pick a single name for the same thing and use it consistently: `base` is called `input_data` in `apply`, `method` is called `scale_method` or `order` elsewhere

tiledb/bioimg/converters/scale.py
```diff
@@ -0,0 +1,272 @@
...
+            ),
+        )
+
+        out_data[tile] = resize(
+            base[tuple(scaled_tile)],
+            tuple(slice_shape),
+            preserve_range=True,
+            order=scale_method,
+            anti_aliasing=True,
+        )
```
gsakkis, line 272: Lots of duplicated (or quadruplicated) code in this module, can be DRY-fied (not necessarily this PR).

tiledb/bioimg/converters/base.py
```diff
@@ -198,7 +209,14 @@ def to_tiledb(
             input_axes = reader.axes
             # Create a TileDB array for each level in range(level_min, reader.level_count)
             uris = []
-            for level in range(level_min, reader.level_count):
+            level_max = reader.level_count if not generate_pyramid else level_min + 1
+            if reader.level_count > 1 and generate_pyramid:
+                print(
+                    "Warning: The image contains multiple levels but pyramid generation is enabled. All levels "
```
gsakkis, line 215: Better `warnings.warn` that print in library code.

gsakkis (changes requested):

tests/integration/converters/test_openslide.py
```diff
@@ -1,3 +1,7 @@
+# import os
+#
+# os.add_dll_directory("C:/openslide/bin")
+
```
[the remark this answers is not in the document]
XanthosXanthopoulos: I will remove it. When we upgrade to python >= 3.8 however we need to address the issue of dll loading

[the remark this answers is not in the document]
gsakkis: That's a quick hack that works only for your Windows local development environment. Committing it to the repo as a comment is certainly not needed. Either downgrade your local development environment to Python 3.7 or just remember to not commit it.

tiledb/bioimg/converters/scale.py
```diff
@@ -12,261 +12,102 @@
...
-    CHUNKED_NON_PROGRESSIVE = 2
-    CHUNKED_PROGRESSIVE = 3
+def _scale_tile(
+    tile: Tuple[slice],
+    base: tiledb.Array,
+    output: tiledb.Array,
+    scale_factors: Sequence[float],
+    order: int,
+) -> None:
+    slice_shape = tuple(dimension.stop - dimension.start for dimension in list(tile))
```
gsakkis, line 29: No need for `list(tile)`, tuples are iterable.

tiledb/bioimg/converters/scale.py
```diff
@@ -12,261 +12,102 @@
...
+    CHUNKED_PROGRESSIVE = auto()
 
 
-class ScalerMode(Enum):
-    NON_PROGRESSIVE = 0
-    PROGRESSIVE = 1
-    CHUNKED_NON_PROGRESSIVE = 2
-    CHUNKED_PROGRESSIVE = 3
+def _scale_tile(
+    tile: Tuple[slice],
```
gsakkis, line 23: This means a tuple of exactly one slice; the correct annotation is `Tuple[slice, ...]`

tiledb/bioimg/converters/base.py
```diff
@@ -343,6 +300,53 @@ def tile_to_tiledb(level_tile: Tuple[slice, ...]) -> None:
...
+    with tiledb.open(base_uri) as a:
+        scaler = Scaler(a, axes, **pyramid_kwargs)
+
+    level = level_min + 1
+
+    for index, dimension in enumerate(scaler.resolutions):
+        uri = os.path.join(output_path, f"l_{level}.tdb")
+        schema = _get_schema(
+            axes=axes,
+            shape=dimension,
```
gsakkis, line 322: Confusing/inconsistent naming: 
- The property is called `resolutions`
- Each element of it is called `dimension`
- This is passed to a parameter called `shape`

Please rename them all to indicate they are shapes (e.g. `resolutions` -> `level_shapes`, `dimension -> shape`).

tiledb/bioimg/converters/scale.py
```diff
@@ -12,261 +12,102 @@
...
+        anti_aliasing=True,
+    )
 
 
 class Scaler(object):
     def __init__(
-        self, base: tiledb.array.DenseArray, axes: Axes, scale_factor: List[int]
+        self,
+        base: tiledb.array,
+        base_axes: Axes,
```
gsakkis, line 51: Change this to `str` for consistency with `scale_axes`

tiledb/bioimg/converters/scale.py
```diff
@@ -12,261 +12,102 @@
...
+        order=order,
+        anti_aliasing=True,
+    )
 
 
 class Scaler(object):
     def __init__(
-        self, base: tiledb.array.DenseArray, axes: Axes, scale_factor: List[int]
+        self,
+        base: tiledb.array,
```
gsakkis, line 50: Replace this with `base_shape: Tuple[int, ...]`; you only need the shape here.

tiledb/bioimg/converters/scale.py
```diff
@@ -12,261 +12,102 @@
...
-                scaled_y_start = tile[self._y_index].start // ratio
-                scaled_y_end = tile[self._y_index].stop // ratio
-                scaled_tile[self._y_index] = slice(scaled_y_start, scaled_y_end)
-
-                tiles.append(tuple(tile))
-
-            self._resolutions.append(tuple(shape))
-            self._tiles.append(tiles)
+            self._scale_factors.append(level_factors)
+            self._scale_factors_progressive.append(level_factors_progressive)
```
gsakkis, line 81: You don't need to compute both `_scale_factors` and `_scale_factors_progressive`; you already know the mode so you can tell which are needed (or neither for non-chunked mode).

tiledb/bioimg/converters/scale.py
```diff
@@ -12,261 +12,102 @@
...
-    NON_PROGRESSIVE = 0
-    PROGRESSIVE = 1
-    CHUNKED_NON_PROGRESSIVE = 2
-    CHUNKED_PROGRESSIVE = 3
+def _scale_tile(
+    tile: Tuple[slice],
+    base: tiledb.Array,
+    output: tiledb.Array,
+    scale_factors: Sequence[float],
+    order: int,
```
gsakkis, line 27: Replace `order` with `**resize_kwargs` and pass them all together (`preserve_range`, `order`, `anti_aliasing`)

tiledb/bioimg/converters/base.py
```diff
@@ -270,59 +268,18 @@ def tile_to_tiledb(level_tile: Tuple[slice, ...]) -> None:
...
-                                if pyramid_mode == ScalerMode.PROGRESSIVE:
-                                    scaler.apply_progressive(r, a, interpolation_method)
-                                else:
-                                    scaler.apply_chunked_progressive(
-                                        r, a, interpolation_method
-                                    )
-                    uris.append(uri)
-                    level += 1
+                uris += _scale(
+                    base_uri=uris[0],
```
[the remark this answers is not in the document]
XanthosXanthopoulos: If level_min is other than 0 but the user wants to generate pyramid level using `uris[level_min]` will result in an `IndexError`

tiledb/bioimg/converters/base.py
```diff
@@ -191,10 +190,16 @@ def to_tiledb(
...
+        :param pyramid_kwargs: Keyword arguments passed to the scaler constructor for generating downsampled versions of
+            the base level.
+            **Note**: Valid keyword arguments are:
+            scale_factors (Required): The downsample factor for each level
+            scale_axes (Required): The axes which will be downsampled
+            chunked (Optional): Default True. If true the image is split into chunks and each one is independently
+            downsampled. If false the entire image is downsampled at once, but it requires more memory.
+            progressive (Optional): Default True. If true each downsampled image is generated using the previous level.
+            If false for every downsampled image the level_min is used, but it requires more memory.
+            order (Optional): Default 1. The interpolation order for downsampling.
```
gsakkis, line 202: Better reuse this more detailed description of the parameter: https://github.com/scikit-image/scikit-image/blob/v0.19.2/skimage/transform/_warps.py#L98-L101

tiledb/bioimg/converters/scale.py
```diff
@@ -35,58 +26,52 @@ def _scale_tile(
...
         scale_axes: str,
-        mode: ScalerMode,
-        order: int,
+        chunked: bool = False,
+        progressive: bool = False,
+        order: int = 1,
     ):
-        self._mode = mode
+        self._chunked = chunked
         self._order = order
```
gsakkis, line 44: Better save all the resize kwargs since they are referenced more than once: 
`self._resize_kwargs = dict(order=order, preserve_range=True, anti_aliasing=True)`

tiledb/bioimg/converters/scale.py
```diff
@@ -96,18 +81,15 @@ def apply(self, base: tiledb.Array, output: tiledb.Array, level: int) -> None:
             )
         else:
             with Pool(multiprocessing.cpu_count()) as pool:
```
gsakkis, line 83: 1. No need for `multiprocessing.cpu_count()`, that's the default
2. The [`concurrent.futures`](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor-example) module is slightly preferred to `multiprocessing` these days. Since you don't need the results of the parallelized function, you can [submit](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.submit) each callable to be scheduled and finally [wait](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.wait) for all of them to finish.

gsakkis (changes requested):
Hopefully last batch of comments :crossed_fingers:

tiledb/bioimg/converters/scale.py
gsakkis: Another idea: add an optional `max_workers` parameter (default=`multiprocessing.cpu_count()`) in `pyramid_kwargs` just like the one in `ImageConverter.to_tiledb` to allow more flexibility. May also allow `max_workers=0` to mean sequential scaling, one tile at a time.

gsakkis (approved):
LGTM, great work! :100:

PR #62, opened 2023-01-24
Adding group metadata mapping

ktsitsi:
This PR:

- Adds a `dataset_type` key in group's metadata for recognising `bioimg` groups.  
- Adds a `kv` mapping in group metadata where `key: level_uri` and `value: [level_id, level_shape, downsample_factors]`
- Moves some global constants in package's `__init__` file


gsakkis:
Sample group metadata after this PR:
```
{'C3L-00445-22.tdb/l_2.tdb': '[2, [3, 2072, 5851], [1.0, 1.0, 1.0]]',
 'C3L-00445-22.tdb/l_3.tdb': '[3, [3, 1036, 2925], [1.0, 2.0, 2.0003418803418804]]',
 'axes': 'YXC',
 'dataset_type': 'BIOIMG',
 'fmt_version': 1,
 'json_tiffwriter_kwargs': '{"bigtiff": false, "byteorder": "<", "append": true, "imagej": false, "ome": false}',
 'pkg_version': '0.1.1.dev21+dirty'}
```

Several issues:
- The entries for each level are injected in the top level metadata instead of being grouped together under a single key. There is no (easy) way for a client to say "get the metadata of level 1"
- Each entry is a list instead of dict; `'[2, [3, 2072, 5851], [1.0, 1.0, 1.0]]` makes no sense without labels.
- The second element is the `level_shape`, which isn't necessarily compatible with the original `axes`. We should also save the canonical axes for each level (`level_axes`).
- The third element is wrong: the downsampling factor for each level is a single value, not a list/array. See `TileDBOpenSlide.level_downsamples` for how to compute it (and ideally reuse the same code).

Expected metadata:
```
{'axes': 'YXC',
 'dataset_type': 'BIOIMG',
 'fmt_version': 1,
 'json_tiffwriter_kwargs': '{"bigtiff": false, "byteorder": "<", "append": true, "imagej": false, "ome": false}',
 'levels': '[{"uri": "C3L-00445-22-new.tdb/l_2.tdb", "level": 2, "axes": "CYX", "shape": [3, 2072, 5851], "downsample_factor": 1.0}, {"uri": "C3L-00445-22-new.tdb/l_3.tdb", "level": 3, "axes": "CYX", "shape": [3, 1036, 2925], "downsample_factor": 2.00017094017094}]',
 'pkg_version': '0.1.1.dev21+dirty'}
```

gsakkis (changes requested):
Commented on the proposed metadata format [here](https://github.com/TileDB-Inc/TileDB-BioImaging/pull/62#issuecomment-1404722133). A unit test to validate it would be nice.

tests/integration/converters/test_ome_zarr.py
```diff
@@ -115,3 +116,21 @@ def test_ome_zarr_converter_incremental(tmp_path):
     OMEZarrConverter.to_tiledb(input_path, str(tmp_path), level_min=0)
     with TileDBOpenSlide.from_group_uri(str(tmp_path)) as t:
         assert len(tiledb.Group(str(tmp_path))) == t.level_count == 2
+
+
+@pytest.mark.parametrize("series_idx", [0, 1, 2])
+@pytest.mark.parametrize("preserve_axes", [False, True])
+def test_ome_zarr_converter_group_meta(tmp_path, series_idx, preserve_axes):
```
gsakkis, line 123: :100:

tiledb/bioimg/converters/base.py
```diff
@@ -198,6 +198,8 @@ def to_tiledb(
             input_axes = reader.axes
             # Create a TileDB array for each level in range(level_min, reader.level_count)
             uris = []
+            levels_meta = []
+            meta_kvstore: Dict[str, Any] = {}
```
gsakkis, line 202: Redundant, just create a new dict for each level

tiledb/bioimg/converters/base.py
```diff
@@ -227,6 +229,22 @@ def to_tiledb(
                 tiledb.Array.create(uri, schema)
                 uris.append(uri)
 
+                # Calculate downsample factor
+                dims_map = dict(zip([*level_axes.dims], level_shape))
```
gsakkis, line 233: `zip([*level_axes.dims], level_shape)) == zip(level_axes.dims, level_shape))` (strings are iterable, no need to make it a list)

tiledb/bioimg/converters/base.py
```diff
@@ -227,6 +229,22 @@ def to_tiledb(
...
+                    l0_w, l0_h = dims_map["X"], dims_map["Y"]
+                down_factor = (l0_w / dims_map["X"] + l0_h / dims_map["Y"]) / 2.0
+
+                # Store layer mapping with shape value
+                meta_kvstore.update(
+                    uri=uri,
+                    level=level,
+                    axes=level_axes.dims,
+                    shape=level_shape,
+                    downsample_factor=down_factor,
```
gsakkis, line 244: Nit: use the same name for the parameter and the variable

tiledb/bioimg/converters/base.py
```diff
@@ -255,10 +273,12 @@ def tile_to_tiledb(level_tile: Tuple[slice, ...]) -> None:
             # Write group metadata
             with tiledb.Group(output_path, "w") as group:
                 group.meta.update(
-                    reader.group_metadata,
+                    **reader.group_metadata,
```
gsakkis, line 276: No need for `**`, the first (and only) positional parameter to update can be a mapping.

gsakkis (approved):
Minor nits after the changes, LGTM :rocket:

XanthosXanthopoulos:
In case the array is registered in the cloud the uri field holds the name or the cloud uri?

Shelnutt2:
> In case the array is registered in the cloud the uri field holds the name or the cloud uri?

I believe we should not store the `uri` in metadata. What we need is only the group member name. Then we should use the group member API (i.e group["level_0"].uri) to get the uri.

PR #64, opened 2023-01-25
Allow custom compressor 

gsakkis:
First PR (out of 2) for webP support:
- Introduce optional `compressor` parameter in `ImageConverter.to_tiledb`. Defaults to the currently hardcoded value (`tiledb.ZstdFilter(level=0)`).
- Refactor code to pave the way for adding WebP. In particular, add a `get_pixel_depth()` function (better name suggestions are welcome) that returns the number of values encoded for each pixel. For non-webp compressors (i.e. this PR), this is always 1. For webP, the pixel depth will be determined by the selected colorspace format as described [here](https://github.com/TileDB-Inc/TileDB/blob/master/format_spec/filters/webp.md).
- Mostly unrelated but useful change: move the bulk of the `TileDBOpenSlide` functionality to a new `TileDBOpenSlideLevel` class and make the former essentially a sequence of the latter.

### DIFF
--- a/tiledb/bioimg/openslide.py
+++ b/tiledb/bioimg/openslide.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import warnings
-from typing import Any, Iterator, Mapping, Sequence, Tuple, Union
+from typing import Any, Iterator, Mapping, MutableMapping, Sequence, Tuple, Union
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


tiledb/bioimg/openslide.py
```diff
@@ -143,24 +139,64 @@ def get_best_level_for_downsample(self, factor: float) -> int:
...
-        source_axes = Axes(dims)
-        target_axes = self.original_axes if to_original_axes else Axes("YXC")
-        return AxesMapper(source_axes, target_axes).map_array(array[selector])
+        return axes_mapper.map_array(array[selector])
+
+    def close(self) -> None:
+        self._tdb.close()
+
+
+def get_pixel_depth(obj: Union[tiledb.Array, tiledb.Filter]) -> int:
```
XanthosXanthopoulos, line 197: Pixel depth is a property of the original image and should be extracted from the original file

gsakkis: Indeed; putting this on draft until we get the channel order data from the original file.

ktsitsi (approved):

PR #65, opened 2023-01-25
Add WebP support

gsakkis:
Follow-up to https://github.com/TileDB-Inc/TileDB-BioImaging/pull/64 for enabling `tiledb.WebpFilter` support.

In an ideal world, this PR (actually 4906742477adfb53152c7a31884ade9755bb426a) wouldn't be necessary: https://github.com/TileDB-Inc/TileDB-BioImaging/pull/64 already allows any TileDB filter to be passed through. Unlike most(all?) other filters that are agnostic with respect to the shape of the array however, `WebpFilter` is [special](https://github.com/TileDB-Inc/TileDB-Py/pull/1564/files): 
- It expects the array to provide two dimensions for Y, X pixel position.
- Dimensions may be defined with any name, but Y, X should be at dimension index 0, 1 respectively.
- Even worse, the (nominally) 2D  `YX` array is semantically 3D `YXC`: what `WebpFilter` calls `X` is actually `X * C`, the cartesian product of width with the number of channels (aka [pixel depth](https://github.com/TileDB-Inc/TileDB/blob/dev/format_spec/filters/webp.md)). 

4906742477adfb53152c7a31884ade9755bb426a tries to abstract away the `2D`:left_right_arrow:`3D` back and forth inside the `ToWebPAxesMapper` and `FromWebPAxesMapper` subclasses of `AxesMapper`, leaving most of the remaining code blissfully unaware of the underlying complexity. Thus any 3D image with `YXC` axes (or any permutation thereof) can be transparently written to (and read from) a 2D TileDB array with WebP compression.

Other notes:
- Set as draft until [this fix](https://github.com/TileDB-Inc/TileDB-Py/pull/1586) is released.
- Currently only the `RGB` and `RGBA`  [input formats](https://github.com/TileDB-Inc/TileDB/blob/dev/format_spec/filters/webp.md#input-and-output-layout) are supported, and the user is expected to specify the correct format based on the source image. In the future it should be possible to have the input format selected automatically based on the available channels of the source image.


tests/integration/converters/test_ome_tiff.py
```diff
@@ -44,6 +45,49 @@ def test_ome_tiff_converter(tmp_path, open_fileobj, preserve_axes):
...
+        assert isinstance(region, np.ndarray)
+        assert region.ndim == 3
+        assert region.dtype == np.uint8
+        img = PIL.Image.fromarray(region)
+        assert img.size == (300, 400)
+
+        for level in range(t.level_count):
+            region_data = t.read_region((0, 0), level, t.level_dimensions[level])
+            level_data = t.read_level(level)
+            np.testing.assert_array_equal(region_data, level_data)
```
XanthosXanthopoulos, line 88: For lossy compression options this test will always fail. Even using `numpy.testing.assert_allclose` can fail in case the image contains sharp variations.

gsakkis: Addressed at https://github.com/TileDB-Inc/TileDB-BioImaging/pull/65/commits/3b06648abdf2df716ebc9ed7ff70e8b49ab49aed using `skimage.metrics.structural_similarity` that you introduced at https://github.com/TileDB-Inc/TileDB-BioImaging/pull/61.

PR #66, opened 2023-01-31
[HOTFIX] Changes group metadata uris from paths to member names

ktsitsi:


tiledb/bioimg/converters/base.py
```diff
@@ -227,7 +227,7 @@ def to_tiledb(
 
                 # Store layer mapping with shape value
                 meta_kvstore = {
-                    "uri": uri,
+                    "uri": os.path.basename(uri),
```
gsakkis, line 230: - Key should be renamed to to "name"
- `os.path.basename(uri)` is also referenced later in `group.add()`, better assign it to a variable.

gsakkis (commented):
LGTM but should be postponed because https://github.com/TileDB-Inc/TileDB-BioImaging/pull/61/ has duplicated this logic. So
1. Merge https://github.com/TileDB-Inc/TileDB-BioImaging/pull/61/
2. Refactor base.py to DRY-fy it (I'm working on it atm)
3. Rebase/update this PR

tiledb/bioimg/converters/base.py
ktsitsi: For the second point `group.add()` is in an external `for` loop so there are no related. Reverted the change cause the tests were failing.

gsakkis:
Subsumed by https://github.com/TileDB-Inc/TileDB-BioImaging/pull/68

PR #73, opened 2023-02-07
[Feature] Adding non-empty name to data attribute

ktsitsi:
This PR:

- Changes attribute name from 'empty' to `intensity` following this [wiki](https://en.wikipedia.org/wiki/Channel_(digital_image)) source

> If the RGB image is 24-bit (the industry standard as of 2005), each channel has 8 bits, for red, green, and blue—in other words, the image is composed of three images (one for each channel), where each image can store discrete pixels with conventional brightness intensities between 0 and 255.

**Warning**: This change breaks backwards compatibility and alters format specification


tiledb/bioimg/converters/base.py
```diff
@@ -392,7 +392,7 @@ def _create_image_pyramid(
     compressor: tiledb.Filter,
     pyramid_kwargs: Mapping[str, Any],
 ) -> None:
-    with tiledb.open(base_uri, "r") as a:
+    with tiledb.open(base_uri, attr="intensity") as a:
```
XanthosXanthopoulos, line 395: Is the specification of the attribute name necessary when opening an array for either read or write since we only have one attribute? Because in `line 260` the array is opened without that.

ktsitsi: In `L:260` the array is opened in `mode=w` so `attr` is not necessary there. The reason why we introduce this approach in `mode=r` is cause even with one attribute arrays the resulted slices of arrays are of type `ordered_dict`, if `attr` is not set during opening. If set though we get `numpy.arrays` as a return value and thus the introduction of the `attr` arg during opening does not cause too many code changes.

gsakkis: Not sure if it's necessary for writes as well but in either case it's very unlikely that is not needed just in line 260. 

A good reminder of the value of [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself): `attr="intensity"` is repeated 6 times, plus a most-likely missing 7th. This **must** be DRY-fied.

gsakkis (changes requested):

tiledb/bioimg/converters/base.py
ktsitsi: It is not needed in `L:409` as well, it is typo. No `write_op` cares about the array being opened with the `attr` value, since the `attr` value is given during the creation of the array schema. I get that in read mode we could dry-fy it but this would require a wrapper of `tiledb.open` which idk if it is the recommended way to go.

XanthosXanthopoulos: A global read_kwargs wouldn't suffice? It would be defined once so changes would be in a single point but it should be referenced in every location we read.

gsakkis: At the very least the hardcoded `"intensity"` should be replaced with a global var (e.g. `ATTR_NAME`) but IMO a `tiledb.open` wrapper makes sense here so that any reader/writer of the code shouldn't have to know or guess whether `attr` is necessary or not based on `mode`.

tiledb/bioimg/openslide.py
```diff
@@ -152,7 +152,7 @@ def _read_image(
 
 class TileDBOpenSlideLevel:
     def __init__(self, uri: str):
-        self._tdb = tiledb.open(uri)
+        self._tdb = tiledb.open(uri, attr="intensity")
```
XanthosXanthopoulos, line 155: Needs to be replaced with the global variable

XanthosXanthopoulos (approved):

tiledb/bioimg/openslide.py
gsakkis: Being ideologically against `DRY` and failing to use `grep` (or similar) is recipe for trouble  :wink:

```
$ ag intensity -ws

tiledb/bioimg/openslide.py
155:        self._tdb = tiledb.open(uri, attr="intensity")

tiledb/bioimg/converters/__init__.py
3:ATTR_NAME = "intensity"

tiledb/bioimg/converters/base.py
350:    attr = tiledb.Attr(name="intensity", dtype=attr_dtype, filters=[compressor])

tests/__init__.py
25:                name="intensity",
```

gsakkis: Half a dozen more broken `tiledb.open` calls in tests. Good(?) thing we open the array just for inspecting the schema and don't read any data:
```
$ ag tiledb.open tests 

tests/integration/converters/test_openslide.py
30:    with tiledb.open(str(tmp_path / "l_0.tdb")) as A:

tests/integration/converters/test_ome_tiff.py
32:            with tiledb.open(str(tmp_path / f"l_{i}.tdb")) as A:
53:    with tiledb.open(str(tmp_path / "l_0.tdb")) as A:
56:    with tiledb.open(str(tmp_path / "l_1.tdb")) as A:
59:    with tiledb.open(str(tmp_path / "l_2.tdb")) as A:
164:    with tiledb.open(str(tiledb_path / "l_0.tdb")) as A:

tests/integration/converters/test_ome_zarr.py
25:    with tiledb.open(str(tmp_path / "l_0.tdb")) as A:

tests/unit/test_openslide.py
21:                with tiledb.open(level_path, "w") as A:
```

tiledb/bioimg/__init__.py
```diff
@@ -0,0 +1,10 @@
+from typing import Any, Optional
+
+import tiledb
+
+ATTR_NAME = "intensity"
+
+
+def open(uri: str, *args: Optional[Any], **kwargs: Optional[Any]) -> tiledb.Array:
```
gsakkis, line 8: 1. This is an internal helper function, probably shouldn't live in the main public entry point of the package. I'd put it in `base.py` but this creates a cyclic import with `openslide.py` so perhaps a new `utils.py`? As a bonus, a new module could be used to move there all the private functions of `base.py` and leave just the base classes.
2. The original `tiledb.open` signature doesn't have to be preserved, it can be simplified to `(uri, mode="r")`.

ktsitsi: 1. I moved `open` and some funcs from `base` module to a newly created under `bioimg`. However one of them creates again cyclic imports so I left it under `base`. Ideally there should have been 2 utils modules under `converters` package where we would store all the mangled funcs of the base module without cyclic imports and one under `bioimg` that would host `_open` which is used both from `converters` modules but also `openslide`.

2. Is it bad if it is preserved?

ktsitsi: 1. I moved open and some funcs from base module to a newly created helpers.py under bioimg. However one of them creates again cyclic imports so I left it under base. Ideally there should have been 2 utils modules one under converters package where we would store all the mangled funcs of the base module without cyclic imports and one under bioimg that would host _open which is used both from converters modules but also openslide.

2. Is it bad if it is preserved?

tiledb/bioimg/converters/base.py
```diff
@@ -22,6 +20,14 @@
...
 from tiledb.cc import WebpInputFormat
 
+from ..helpers import (
+    _create_image_pyramid,
+    _get_pixel_depth,
+    _get_schema,
+    _iter_levels_meta,
+    _open,
+    _ReadWriteGroup,
+)
```
gsakkis, line 30: The convention is that underscored names are not exported from the module, so all these names should lose the leading underscore now.

Some people make the module name itself "private" (i.e. `_helpers`). Not too keen on it (most modules are usually "private" from the POV of the package user) but not strongly against it either.

gsakkis (approved):
One last comment but LGTM :rocket:

tiledb/bioimg/converters/base.py
ktsitsi: Also the `open` function if loses its leading underscore should change name as well as it shadows the built in `open`.

gsakkis: Yeap, candidate names off the top of my head: `tiledb_open`, `tdb_open`, `open_level_array`, `open_bioimg_array`, ...

ktsitsi: How about `open_bioimg`? I like the best `open_bioimg_array `, but I think it is too long so I cropped it. I think that for internal use it is safe to assume that we open an `array`.

PR #74, opened 2023-02-08
Xan/sc 24385/visualization helpers

XanthosXanthopoulos:
This PR implements a helper function that uses the group metadata to generate the metadata required for the visualization layer.


ktsitsi (commented):
Some minor. LGTM otherwise.

tiledb/bioimg/visualization/helpers.py
```diff
@@ -0,0 +1,65 @@
+import json
+from dataclasses import dataclass
+from typing import List, Optional, Sequence
+
+import tiledb
+
+from ..converters.axes import Axes
```
gsakkis, line 7: Will there be (a lot) more stuff under `visualization` in the near to mid-term future? If not I'd keep it in a `visualization.py` module until it grows enough to justify a subpackage. 

As an aside, I think the `converters` subpackage was a bad call in retrospect given the coupling between the modules: IMO `from ..converters.axes import Axes` is a code structure smell, and so are these:
```py
# tiledb.bioimg.openslide
from .converters.axes import Axes

# tiledb.bioimg.converters.base
from ..helpers import ...
from ..openslide import ...
```

tiledb/bioimg/visualization/helpers.py
```diff
@@ -0,0 +1,65 @@
...
+    width: int
+    height: int
+    namespace: str
+    array_uri: str
+    axes: str
+
+
+def calculate_metadata(
+    namespace: str, group_uri: str, config: Optional[tiledb.Config] = None
+) -> Sequence[ZoomLevelRecord]:
```
gsakkis, line 24: `Metadata` is too overloaded and generic term in this project. I'd rename it to (say) `iter_zoom_level_records` and change the return type to `Iterator[ZoomLevelRecord]`.

tiledb/bioimg/visualization/helpers.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Calculates the required metadata for the Tile Image viewer.
+
+    :param namespace: the namespace of the group
+    :param group_uri: path to the group
+    :param config: the configuration for constructing the tileDB context
+    """
+    metadata: List[ZoomLevelRecord] = []
+
+    context = None if config is None else tiledb.Ctx(config)
```
gsakkis, line 34: Redundant, `tiledb.Ctx(config)` works even when `config is None`.

tiledb/bioimg/visualization/helpers.py
```diff
@@ -0,0 +1,65 @@
...
+    group = tiledb.Group(group_uri, "r", ctx=context)
```
gsakkis, line 35: 1. Open the group as context manager and drop the `group.close()` at the end: `with tiledb.Group(...) as group: ...`
2. [Nit] `mode="r"` is more obvious that a "r" positional parameter.

tiledb/bioimg/visualization/helpers.py
```diff
@@ -0,0 +1,65 @@
...
+    zoom_level = 0
+
+    for meta in json.loads(group.meta["levels"])[::-1]:
```
gsakkis, line 38: 1. `reversed(foo)` is more readable than `foo[::-1]` (and in theory more efficient for large sequences)
2. It's not (immediately) obvious why this has to be done in reversed level order, a comment would be nice.

tiledb/bioimg/visualization/helpers.py
```diff
@@ -0,0 +1,65 @@
...
+        axes = Axes(meta["axes"])
+        axes_mapper = axes.mapper(Axes("XYC"))
+        shape = axes_mapper.map_shape(meta["shape"])
+
+        zoom_factor = round(shape[0] / metadata[-1].width) if len(metadata) > 0 else 2
```
gsakkis, line 43: 1. A `width, height, channel = shape` unpacking would make things a bit clearer.
2. `if len(metadata) > 0` is more idiomatically spelled as `if metadata`

tiledb/bioimg/visualization/helpers.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if not ((zoom_factor & (zoom_factor - 1) == 0) and zoom_factor != 0):
```
gsakkis, line 45: Every time (thankfully rarely) I see a bitwise `&` or `|` I have to spend several seconds/minutes to figure out:
- (a) if this was really meant to be logical `and` or `or`
- (b) or if it is intentionally bitwise operator, and if yes why
- (c) if it [makes a difference](https://stackoverflow.com/questions/3845018/boolean-operators-vs-bitwise-operators) in this case or not:
```py
In [1]: 4 & 0, 4 and 0
Out[1]: (0, 0)

In [2]: 4 & 1, 4 and 1
Out[2]: (0, 1)
```

TL;DR: Do **not** use bitwise `and/or` without a very good reason, and if there is a very good reason (e.g. bit fiddling) please add a comment or two.

tiledb/bioimg/visualization/helpers.py
```diff
@@ -0,0 +1,65 @@
...
+            raise ValueError("Only power of two zoom factors are supported")
+
+        for i in range(bin(zoom_factor - 1).count("1")):
```
gsakkis, line 48: Another case a reader has to spend time and mental effort to figure out what this expression does and why, not to mention the inefficiency of converting an int to string and count its 1s:
 - What does the number of 1s in the binary representation of `zoom_factor-1` have to do with anything?
- Why `zoom_factor-1` (and not, say, `zoom_factor`)?

I'm not saying it's wrong, I'm just saying I have to spend several minutes to figure it out.

tiledb/bioimg/visualization/helpers.py
```diff
@@ -0,0 +1,65 @@
...
+            metadata.append(
+                ZoomLevelRecord(
+                    zoom_level=zoom_level,
+                    image_level=meta["level"],
+                    downsample=2 ** (bin(zoom_factor - 1).count("1") - i - 1),
```
gsakkis, line 53: As if the obfuscated inefficient expression above wasn't enough, (a) it is repeated here and (b) is needlessly recomputed on every iteration of the loop.

gsakkis:
More readable version for the downsample logic:
```diff
@@ -2,6 +2,8 @@ import json
 from dataclasses import dataclass
 from typing import List, Optional, Sequence
 
+import numpy as np
+
 import tiledb
 
 from ..converters.axes import Axes
@@ -42,15 +44,16 @@ def calculate_metadata(
 
         zoom_factor = round(shape[0] / metadata[-1].width) if len(metadata) > 0 else 2
 
-        if not ((zoom_factor & (zoom_factor - 1) == 0) and zoom_factor != 0):
+        n = np.log2(zoom_factor)
+        if n != n.astype(int):
             raise ValueError("Only power of two zoom factors are supported")
 
-        for i in range(bin(zoom_factor - 1).count("1")):
+        for downsample in reversed(2 ** np.arange(n.astype(int))):
             metadata.append(
                 ZoomLevelRecord(
                     zoom_level=zoom_level,
                     image_level=meta["level"],
-                    downsample=2 ** (bin(zoom_factor - 1).count("1") - i - 1),
+                    downsample=downsample,
                     width=shape[0],
                     height=shape[1],
                     namespace=namespace,
```

tiledb/bioimg/visualization/helpers.py
gsakkis: It should be possible to wrap a generator into a `list()` before passing it, no?

gsakkis (approved):
LGTM :shipit:

PR #75, opened 2023-02-09
Allow passing ImageReader to `ImageConverter.to_tiledb`

gsakkis:
Allow passing an existing `ImageReader` instance as `input_path` to `ImageConverter.to_tiledb`. 
Uses cases:
  - Not having to create a `FooConverter` class for every `FooReader`; a user can call `ImageConverter.to_tiledb(FooReader(...), ...)` instead.
  - In case one needs to convert the same input more than once (e.g. to both lossless and lossy output) and creating a `FooReader` is expensive, create only one instance and pass it to multiple `to_tiledb` calls.

Also some more `ImageConverter.to_tiledb` refactoring  while at it.

**Breaking Backwards Compatibility**: Breaking API `to_tiledb` with `source` argument instead of `input_path`

### DIFF
--- a/tiledb/bioimg/converters/base.py
+++ b/tiledb/bioimg/converters/base.py
@@ -31,7 +31,7 @@ from ..helpers import (
 from ..openslide import TileDBOpenSlide
 from ..version import version as PKG_VERSION
 from . import DATASET_TYPE, FMT_VERSION
-from .axes import Axes, AxesMapper
+from .axes import Axes
 from .tiles import iter_tiles, num_tiles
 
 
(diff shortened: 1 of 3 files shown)


tiledb/bioimg/converters/base.py
```diff
@@ -164,7 +164,7 @@ def from_tiledb(
     @classmethod
     def to_tiledb(
         cls,
-        input_path: str,
+        input_path: Union[str, ImageReader],
```
ktsitsi, line 167: If this is not a `path` anymore let's rename it. At the top of my head just `input` would make sense.

tiledb/bioimg/converters/base.py
```diff
@@ -213,15 +213,17 @@ def to_tiledb(
                 chunked downsampling. If None, it will default to the number of processors
                 on the machine, multiplied by 5.
         """
-        if cls._ImageReaderType is None:
+        if isinstance(input_path, ImageReader):
```
ktsitsi, line 216: In case of misuse doesn't this allow a Frankenstein expression of `OmeZarrConverter.to_tiledb(OmeTiffReader(...),...)` ?

tiledb/bioimg/converters/base.py
```diff
@@ -213,15 +213,17 @@ def to_tiledb(
...
+            reader = input_path
+        elif cls._ImageReaderType is not None:
+            reader = cls._ImageReaderType(input_path, **reader_kwargs)
+        else:
```
ktsitsi, line 220: This new functionality is not tested, could we have some tests?

ktsitsi (changes requested):

tiledb/bioimg/converters/base.py
gsakkis: Good point, currently this works and just ignores the class. It is possible to detect this case and raise an error, if desired.

XanthosXanthopoulos (approved):
LGTM 🚀

PR #76, opened 2023-02-10
Fix: save non-tiledb-cloud URIs as relative

gsakkis:
Fix regression in adding local array URIs.

### DIFF
--- a/tiledb/bioimg/helpers.py
+++ b/tiledb/bioimg/helpers.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
-import os
+from pathlib import Path
 from typing import Any, Iterator, Mapping, Optional, Tuple
-from urllib.parse import urlparse
+from urllib.parse import urljoin, urlparse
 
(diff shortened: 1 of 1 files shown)


tiledb/bioimg/helpers.py
```diff
@@ -32,7 +34,7 @@ def get_or_create(self, name: str, schema: tiledb.ArraySchema) -> Tuple[str, boo
         if name in self.r_group:
             uri = self.r_group[name].uri
         else:
-            uri = os.path.join(self.r_group.uri, name)
+            uri = os.path.join(self._uri, name)
```
XanthosXanthopoulos, line 37: `os.path.join` will produce platform dependent paths where might be incompatible with the remote storage. An example of that is if a local file on a windows machine ingests a file and saves it in an s3 bucket. Then the array uri contains mixed `/` and `\`

ktsitsi (approved):
LGTM!

PR #78, opened 2023-02-17
Added support for Windows-style paths

XanthosXanthopoulos:
With `urllib` when joining paths that started without http (or https) the base path was discarded resulting in wrong paths. By using `os.path.join` and normalizing all path separators with `/` we can join any path under the assumption that we only have one root (or absolute) path.

### DIFF
--- a/tiledb/bioimg/helpers.py
+++ b/tiledb/bioimg/helpers.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
+import os
 from pathlib import Path
 from typing import Any, Iterator, Mapping, Tuple
-from urllib.parse import urljoin, urlparse
+from urllib.parse import urlparse
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


tiledb/bioimg/helpers.py
```diff
@@ -17,8 +18,11 @@ class ReadWriteGroup:
     def __init__(self, uri: str):
         parsed_uri = urlparse(uri)
         # normalize uri if it's a local path (e.g. ../..foo/bar)
-        if parsed_uri.scheme in ("", "file"):
-            uri = str(Path(parsed_uri.path).resolve())
+
+        # Windows paths produce single letter scheme matching the drive letter
+        # Unix absolute path produce an empty scheme
+        if len(parsed_uri.scheme) < 2 or parsed_uri.scheme == "file":
```
ktsitsi, line 24: Generally magic numbers should be avoided, but since we will revisit the whole path resolution in future ticket lets leave it as is.

ktsitsi (approved):

PR #80, opened 2023-03-31
Exposing `attr` in open api

ktsitsi:
This PR:

- Exposes an `attr` argument for API calls, thus adds some backwards compatibility for bioimg `arrays` that were created before the new specification where the `attribute` value of array is `intensity`.

### DIFF
--- a/tests/unit/test_openslide.py
+++ b/tests/unit/test_openslide.py
@@ -7,6 +7,8 @@ from tests import get_schema
 from tiledb.bioimg.helpers import open_bioimg
 from tiledb.bioimg.openslide import TileDBOpenSlide
 
+from .. import ATTR_NAME
+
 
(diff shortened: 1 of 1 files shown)


ihnorton (changes requested):
LGTM, just needs a test.

tests/unit/test_openslide.py
```diff
@@ -26,3 +28,7 @@ def r():
         with TileDBOpenSlide(group_path) as t:
             assert t.level_count == len(level_dimensions)
             assert t.level_dimensions == tuple(level_dimensions)
+
```
ihnorton, line 31: Can we also test whether it overriding the name reads correctly?

ihnorton (approved):
Minor question, but LGTM - merge when ready.

PR #81, opened 2023-04-03
Fix double registration

ktsitsi:
This PR:

- Fixes double group registration. The group creation with `tiledb` uri path already registers the group and there is no need for extra registration. It keeps the registration functionality when a uri except for `tiledb` is provided and when the `register_kwargs` are provided.

### DIFF
--- a/tiledb/bioimg/converters/base.py
+++ b/tiledb/bioimg/converters/base.py
@@ -281,8 +281,10 @@ class ImageConverter:
                 ),
             )
 
-        if register_group is not None and urlparse(output_path).scheme == "tiledb":
-            register_group(name=os.path.basename(output_path), **register_kwargs)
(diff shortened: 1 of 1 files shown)


tiledb/bioimg/converters/base.py
```diff
@@ -281,8 +281,10 @@ def to_tiledb(
                 ),
             )
 
-        if register_group is not None and urlparse(output_path).scheme == "tiledb":
-            register_group(name=os.path.basename(output_path), **register_kwargs)
+        if register_group is not None and urlparse(output_path).scheme != "tiledb":
```
Shelnutt2, line 284: Is there a reason we don't just do a check on if the group already exists?

Shelnutt2 (approved):

PR #83, opened 2023-04-05
Add windows runner CI

ktsitsi:
This PR:

- Installs openslide from binaries in `win32` runner.
- Adds a `win32` CI. Now tests run in both `linux` and `win32`


tests/integration/converters/test_ome_tiff.py
```diff
@@ -22,6 +22,7 @@ def test_ome_tiff_converter(tmp_path, open_fileobj):
         with open(input_path, "rb") as f:
             OMETiffConverter.to_tiledb(f, output_path)
     else:
+        print(output_path)
```
ihnorton, line 25: Cruft?

ihnorton:
Thanks

PR #84, opened 2023-05-04
Excluding tests and docs from package

ktsitsi:
This PR:

- Fixes the `setup.py` and `setup.cfg` files s.t. to exclude `tests*` and `docs*` from package distribution.


ihnorton (commented):
Why do we need both manifest and setup.cfg entires?

ihnorton (approved):

PR #87, opened 2023-05-17
Support different compressors per level

ktsitsi:
This PR:

- Enables the user to ingest a multi-layer image by compressing each level with a separate filter.
- The default compression filter remains to be `Zstd` as it used to be.
- The user can either select a filter to be applied to all levels of the image `or` provide a mapping of levels -> filter for each image
- In the case of scaling an image with our `Scaler`, the newly constructed scaled levels are being compressed/filtered with the filter of the `base_layer`
-  In the metadata section due to this feature we have a change as now the `pixel_depth` value can differ among the different levels as it is a function of the filter that is used. We store this information in the group metadata as a Mapping as well.
- The changes were done in a way that the API of the `to_tiledb` does not change.
- Possibly alongside with some other changes that are happening in other PRs like `channel metadata` we will have to update the `FMT_VERSION` when released.

### DIFF
--- a/tests/integration/converters/test_ome_tiff.py
+++ b/tests/integration/converters/test_ome_tiff.py
@@ -1,3 +1,4 @@
+import itertools
 import json
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


ihnorton (changes requested):
Could you please add some tests?

ihnorton (approved):
👍 thanks

PR #89, opened 2023-05-26
Adding a from_bioimg wrapper for converters

ktsitsi:
This PR:

- creates a `from_bioimg` function wrapper for supporting under one command all the converters
- creates an `enum` structure with the `converters` options. For now the converters argument is mandatory.

```python
import tiledb.bioimg

tiledb.bioimg.from_bioimg(src:str, dest:str, converter: tiledb.bioimg.Converter)

```


tiledb/bioimg/wrappers.py
```diff
@@ -0,0 +1,23 @@
...
+) -> Type[ImageConverter]:
+    if converter is Converters.OMETIFF:
+        OMETiffConverter.to_tiledb(source=src, output_path=dest, **kwargs)
+        return OMETiffConverter
+    elif converter is Converters.OMEZARR:
+        OMEZarrConverter.to_tiledb(source=src, output_path=dest, **kwargs)
+        return OMEZarrConverter
+    else:
+        OpenSlideConverter.to_tiledb(source=src, output_path=dest, **kwargs)
+        return OpenSlideConverter
```
ihnorton, line 23: Can you please make this do the conversion call, instead of only returning the converter?

ktsitsi: Sure that is how I started but it was easier for me to test it this way. I can modify.

PR #90, opened 2023-06-06
Floating point and QPTiff metadata support

XanthosXanthopoulos:
This PR adds support for metadata extraction from QPTiff images (both RGB and multichannel) as well as channel min-max computation from floating point images

### DIFF
--- a/tiledb/bioimg/converters/base.py
+++ b/tiledb/bioimg/converters/base.py
@@ -432,12 +432,15 @@ def _convert_level_to_tiledb(
 
     # Initialize a numpy 2D array to hold the min-max values per channel
     channel_min_max = np.empty((channel_count, 2))
-    channel_min_max[:, 0] = np.repeat(
-        np.iinfo(reader.level_dtype(level)).max, channel_count
-    )
-    channel_min_max[:, 1] = np.repeat(
-        np.iinfo(reader.level_dtype(level)).min, channel_count
-    )
+
+    if np.issubdtype(reader.level_dtype(0), np.integer):
+        min_value = np.iinfo(reader.level_dtype(level)).min
+        max_value = np.iinfo(reader.level_dtype(level)).max
+    else:
+        min_value = np.finfo(reader.level_dtype(level)).min
+        max_value = np.finfo(reader.level_dtype(level)).max
+    channel_min_max[:, 0] = np.repeat(max_value, channel_count)
+    channel_min_max[:, 1] = np.repeat(min_value, channel_count)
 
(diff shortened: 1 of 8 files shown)


tiledb/bioimg/helpers.py
```diff
@@ -169,9 +169,11 @@ def iter_color(attr_type: np.dtype) -> Iterator[Dict[str, int]]:
...
     while True:
         if np.issubdtype(attr_type, np.integer):
-            red = np.random.randint(low=min_val, high=max_val, dtype=attr_type)
-            green = np.random.randint(low=min_val, high=max_val, dtype=attr_type)
-            blue = np.random.randint(low=min_val, high=max_val, dtype=attr_type)
+            red = np.random.randint(low=min_val, high=max_val, dtype=attr_type).item(0)
+            green = np.random.randint(low=min_val, high=max_val, dtype=attr_type).item(
+                0
+            )
+            blue = np.random.randint(low=min_val, high=max_val, dtype=attr_type).item(0)
```
ktsitsi, line 176: Why does this need to change?

tiledb/bioimg/converters/base.py
```diff
@@ -180,16 +186,70 @@ def from_tiledb(
...
+                                    Axes("".join(axes_metadata.get("originalAxes")))
+                                    .webp_mapper(4)
+                                    .inverse
+                                )
+                        else:
+                            axes_mapper = Axes(
+                                "".join(axes_metadata.get("storedAxes"))
+                            ).mapper(Axes("".join(axes_metadata.get("originalAxes"))))
+
+                    dask_array = dask.array.from_tiledb(member.uri, attribute=attr)
```
Shelnutt2, line 238: Why is dask being used here?

XanthosXanthopoulos: It was a temporary fix to export large images locally and to be removed.

Shelnutt2: We cannot introduce dask here for that purpose. We need another solution. Dask dependency is not something we should add just for this. If you want chunked writes we can handle that.

cc: @ihnorton

ktsitsi: We have already a ticket for that, we should plan it and implement it similar to `chunked` ingestion. For now the way we handle the intermediate export with the configurable memory on pods and the bigger memory in local machines will not create problems for most of the files when memory is configured correctly.

ktsitsi (approved):
LGTM! :tada As discussed let's move the reverted changes to a separate PR.

PR #93, opened 2023-06-30
TIFF Random Access Fix

ktsitsi:
This PR:

- Complements the investigation of [sc-30748] and it is a workaround for random access writes of `tifffile` dependency. This fix is OS specific.

### DIFF
--- a/tiledb/bioimg/converters/base.py
+++ b/tiledb/bioimg/converters/base.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import json
-import os.path
+import os
 import warnings
 from abc import ABC, abstractmethod
 from concurrent.futures import ThreadPoolExecutor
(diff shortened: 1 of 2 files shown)


tiledb/bioimg/converters/base.py
```diff
@@ -183,12 +184,9 @@ def from_tiledb(
 
         with tiledb.scope_ctx(config):
             vfs_use = output_path.startswith("s3://")
-            if vfs_use:
-                vfs = tiledb.VFS()
-                destination_uri = vfs.open(output_path, "wb")
-            else:
-                destination_uri = output_path
-
+            destination_uri = (
```
Shelnutt2, line 187: Let's make this a parameter that can be passed from the user. Call it `scratch_space` or similar to distinguish from where the final location it.

Shelnutt2, line 187: Also for the `vfs_use`, lets just use vfs for all cases now. Or we need to use it for azure, gcs, hdfs, s3 at a minimum.

Shelnutt2 (commented):
Nice work, a few comments.

tiledb/bioimg/converters/base.py
ktsitsi: We need to do the latter. The reason for `vfs use` flag is just to exclude local files (`scrach_space`) from be handled by `vfs` which in `tifffile` activates a different write path that uses the VFS IO (`write` especially). So I guess I can inverse the check so all the remote `fs` can be handled the same as `s3` now.

Shelnutt2 (approved):

PR #94, opened 2023-07-05
Xan/ome tiff exporter

XanthosXanthopoulos:


tiledb/bioimg/converters/ome_tiff.py
```diff
@@ -274,17 +286,86 @@ def __init__(self, output_path: str):
         self._output_path = output_path
 
     def write_group_metadata(self, metadata: Mapping[str, Any]) -> None:
-        tiffwriter_kwargs = json.loads(metadata["json_tiffwriter_kwargs"])
-        tiffwriter_kwargs.pop("append")
         self._writer = tifffile.TiffWriter(
-            self._output_path, shaped=False, append=False, **tiffwriter_kwargs
+            self._output_path, shaped=False, bigtiff=True, append=False, ome=True
```
ktsitsi, line 290: Why to overwrite or bypass the `group_metadata`? Aren't these values being read by the `tiff` itself? Why should we hardcode them?

tiledb/bioimg/converters/base.py
```diff
@@ -203,12 +210,24 @@ def from_tiledb(
 
             with slide, writer:
                 writer.write_group_metadata(slide.properties)
-                for level in slide.levels:
-                    if level < level_min:
+                group_metadata = jsonpickle.loads(
```
ktsitsi, line 213: Why do we pass group_metadata into write_level?

tiledb/bioimg/converters/ome_tiff.py
```diff
@@ -274,17 +286,86 @@ def __init__(self, output_path: str):
...
+            write_kwargs["compressionargs"] = {"level": 70}
+            write_kwargs["photometric"] = tiffwriter_metadata.get(
+                "photometric", tifffile.PHOTOMETRIC.RGB
+            )
+            mask[original_axes.index("C")] = slice(0, 3)  # Drop alpha channel
+        else:
+            write_kwargs["compression"] = tifffile.COMPRESSION.ADOBE_DEFLATE
+            write_kwargs["photometric"] = tiffwriter_metadata.get(
+                "photometric", tifffile.PHOTOMETRIC.MINISBLACK
+            )
```
ktsitsi, line 366: Why all this code inflation? Is it necessary to have group metadata in array metadata? Aren't the same for every level?

tiledb/bioimg/helpers.py
```diff
@@ -159,13 +159,16 @@ def get_axes_translation(
...
 
-    yield {"red": max_val, "green": min_val, "blue": min_val, "alpha": max_val}
-    yield {"red": min_val, "green": max_val, "blue": min_val, "alpha": max_val}
-    yield {"red": min_val, "green": min_val, "blue": max_val, "alpha": max_val}
+    if channel_num == 1:
+        yield {"red": max_val, "green": max_val, "blue": max_val, "alpha": max_val}
+    elif channel_num == 3:
+        yield {"red": max_val, "green": min_val, "blue": min_val, "alpha": max_val}
+        yield {"red": min_val, "green": max_val, "blue": min_val, "alpha": max_val}
+        yield {"red": min_val, "green": min_val, "blue": max_val, "alpha": max_val}
```
ktsitsi, line 171: Could you please provide unit tests for `iter_color` it has been modified a lot lately and we want to make sure that we don't patch ad-hoc functions. We need to make sure that the function is complete and functional.

tiledb/bioimg/converters/base.py
```diff
@@ -203,12 +210,24 @@ def from_tiledb(
...
-                    level_metadata = slide.level_properties(level)
-                    writer.write_level_image(level, level_image, level_metadata)
+                    level_image = slide.read_level(idx, to_original_axes=True)
+                    level_metadata = {
+                        "group_metadata": group_metadata,
+                        "writer_metadata": slide.level_properties(idx),
+                    }
+                    writer.write_level_image(
+                        idx,
+                        idx == level_min,
```
ktsitsi, line 226: Why do you need an extra argument in the function `baseline` to perform this condition check when both parts of the condition are being passed into the function?

ktsitsi (changes requested):

tiledb/bioimg/converters/ome_tiff.py
XanthosXanthopoulos: When writing the tiff file we can't estimate the final size whether it will surpass 4GBs so the `bigtiff` is set to true as a foolproof solution. The `ome` is set to true in order to provide a uniform export format from a any input format.

tiledb/bioimg/converters/ome_tiff.py
```diff
@@ -242,10 +254,10 @@ def image_metadata(self) -> Dict[str, Any]:
                     info[key] = value
 
                 if "MPP" in info:
-                    metadata["physicalSizeX"] = metadata["physicalSizeY"] = info.get(
-                        "MPP"
+                    metadata["physicalSizeX"] = metadata["physicalSizeY"] = float(
+                        info.get("MPP", "1")
```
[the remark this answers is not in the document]
XanthosXanthopoulos: It is needed as a float when writing it to the metadata as a decimal otherwise it is in scientific format

tiledb/bioimg/converters/ome_tiff.py
ktsitsi: Gotcha then I am more inclined into hardcoding these values in `group_metadata` property.  Let's dry-fy it.

tiledb/bioimg/converters/ome_tiff.py
```diff
@@ -270,20 +282,112 @@ def original_metadata(self) -> Dict[str, Any]:
...
+    def __init__(self, output_path: str, ome: bool = True):
         self._output_path = output_path
+        self._ome = ome
 
     def write_group_metadata(self, metadata: Mapping[str, Any]) -> None:
-        tiffwriter_kwargs = json.loads(metadata["json_tiffwriter_kwargs"])
-        tiffwriter_kwargs.pop("append")
         self._writer = tifffile.TiffWriter(
-            self._output_path, shaped=False, append=False, **tiffwriter_kwargs
+            self._output_path, shaped=False, bigtiff=True, append=False, ome=self._ome
```
ktsitsi, line 291: Why don't we leave this as is and hardcode the `group_metadata` : 
```python
@property
    def group_metadata(self) -> Dict[str, Any]:
        writer_kwargs = dict(
            bigtiff=self._tiff.is_bigtiff,
            byteorder=self._tiff.byteorder,
            append=self._tiff.is_appendable,
            imagej=self._tiff.is_imagej,
            ome=self._tiff.is_ome,
        )
        return {"json_tiffwriter_kwargs": json.dumps(writer_kwargs)}
```

and after that as it was `self._writer = tifffile.TiffWriter(self._output_path, shaped=False, **tiffwriter_kwargs` calling that will do the job

tiledb/bioimg/converters/ome_tiff.py
```diff
@@ -270,20 +282,112 @@ def original_metadata(self) -> Dict[str, Any]:
...
+        metadata: Mapping[str, Any],
     ) -> None:
-        write_kwargs = json.loads(metadata["json_write_kwargs"])
+        write_kwargs: Dict[str, Any] = dict(metadata)
+        if baseline:
+            write_kwargs["subifds"] = num_levels - 1 if num_levels > 1 else None
+        else:
+            write_kwargs["subfiletype"] = tifffile.FILETYPE.REDUCEDIMAGE
+
+        write_kwargs["tile"] = (256, 256)
```
ktsitsi, line 389: Magic tuple. Define it if it is a constant.

ktsitsi, line 389: @Also why all these cannot be done into `compute_level_metadata` ?

ktsitsi (changes requested):
After these changes LGTM ! Thanks! :star:

tiledb/bioimg/converters/ome_tiff.py
XanthosXanthopoulos: I am more inclined to store the actual tiff metadata and choose not to use them during export rather than storing a set of "ideal metadata." Having the metadata can provide debug info in case of erroneous ingestion. Also, ingestion and export are being decoupled since no OS/platform specific options (like byteorder) are shared

ktsitsi (approved):

PR #98, opened 2023-08-10
Add io separate configs for exportation

ktsitsi:
This PR:

- Adds two separate config args for exportation in case the destination has other access credentials needs than the source.

### DIFF
--- a/tiledb/bioimg/converters/base.py
+++ b/tiledb/bioimg/converters/base.py
@@ -194,7 +194,8 @@ class ImageConverter:
         *,
         level_min: int = 0,
         attr: str = ATTR_NAME,
-        config: Union[tiledb.Config, Mapping[str, Any]] = None,
+        in_config: Union[tiledb.Config, Mapping[str, Any]] = None,
+        out_config: Union[tiledb.Config, Mapping[str, Any]] = None,
         scratch_space: str = DEFAULT_SCRATCH_SPACE,
         **writer_kwargs: Mapping[str, Any],
     ) -> Type[ImageConverter]:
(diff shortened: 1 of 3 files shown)


tiledb/bioimg/converters/base.py
```diff
@@ -194,7 +194,8 @@ def from_tiledb(
         *,
         level_min: int = 0,
         attr: str = ATTR_NAME,
-        config: Union[tiledb.Config, Mapping[str, Any]] = None,
+        in_config: Union[tiledb.Config, Mapping[str, Any]] = None,
```
Shelnutt2, line 197: We need to keep `config` for backwards comaptibility.

tiledb/bioimg/converters/base.py
```diff
@@ -205,51 +206,50 @@ def from_tiledb(
...
-                    writer.write_level_image(
-                        level_image,
-                        level_metadata,
-                    )
 
             if vfs_use:
                 # Flush to remote
                 with open(destination_uri, "rb") as data:
-                    vfs = tiledb.VFS()
+                    vfs = tiledb.VFS(config=output_config)
```
Shelnutt2, line 252: I think we need some condition to determine to use `output_config` or just `config`. If the user only pass `config` (deprecated but supported) won't we ignore it now?

ktsitsi: I see the following scenarios:
1. **The user uses an older version, where we do not support separate configuration for the exported path.** (👍 )
2. **The user uses the updated version, where he wants to export from remote source to local output path**, thus uses only `config` arg. The `vfs_use` is becoming `False` as the output path is local and the `destination_uri` gets the local value from here 
```python
destination_uri = (
            os.path.join(scratch_space, os.path.basename(out_uri_res))
            if vfs_use
            else out_uri_res
        )
```
and the writer following writes directly to this local output path. ( 👍 )
3. **The user wants to export to a remote and the source is also remote.** Then normally he should use one `config` for source and one `output_config` for the destination  if the configs are different, while also providing a remote `output_path`. The read will happen and the intermediate file will be stored in the `scratch space`, and now what happens next is upon us. In this implementation if the `output_config` is not given and as said the `output_path` is also remote we fail in the `vfs.open` where the `config=output_config=None` we will not be able to access the remote destination path due to missing credentials. Now the question in semantics is if we want to silently assign `config` to `output_config` if the latter is not defined as a default behaviour, thinking that the source and dest are under the same access credentials policy. (👍 )
4. For the local cases for remote and dest should work fine with `configs=None`(👍 )

Is there any case which I missed?

ktsitsi: So given that with current version user does not have the ability to use different configs, let's follow the semantics I described earlier concerning the following.

>  Now the question in semantics is if we want to silently assign config to output_config if the latter is not defined as a default behaviour, thinking that the source and dest are under the same access credentials policy. (👍 )

This way the user that upgrades and uses only config without knowing the existence of the new features will be able to export to a remote source, but he/she is limited by the package version, where both the source and destination should comply to the same credential access. Would it be sufficient for compatibility?

PR #100, opened 2023-08-11
Migrate to mamba-org/setup-micromamba@v1

jdblischak:
mamba-org/provision-with-micromamba is deprecated (https://github.com/mamba-org/provision-with-micromamba/pull/122)

Unfortunately it looks like selectors are no longer supported, so I installed `openslide` in a separate step

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,19 +21,20 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Install Conda environment with Micromamba
-      uses: mamba-org/provision-with-micromamba@main
+      uses: mamba-org/setup-micromamba@v1
       with:
-        environment-file: false
         environment-name: test
-        channels: conda-forge
         cache-downloads: true
-        extra-specs: |
+        create-args: >-
           python=3.7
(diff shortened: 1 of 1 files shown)


jdblischak:
The [jobs](https://github.com/jdblischak/TileDB-BioImaging/actions/runs/5836983215) passed on my fork

Shelnutt2 (approved):
Thanks!

jdblischak:
I was mistaken. The [second attempt also failed](https://github.com/TileDB-Inc/TileDB-BioImaging/actions/runs/5847084080/attempts/2)

Turns out this is a known Issue (#99), and the plan is to remove the Windows CI

ktsitsi:
> I was mistaken. The [second attempt also failed](https://github.com/TileDB-Inc/TileDB-BioImaging/actions/runs/5847084080/attempts/2)
> 
> Turns out this is a known Issue (#99), and the plan is to remove the Windows CI

Hi @jdblischak for now I have left this un-merged since ideally we would like to keep windows in CI, instead of removing it we might want to increase the runner's lifespan duration for supporting the slower tests in Windows.

PR #101, opened 2023-08-17
No-op and warning when incremental ingestion with incompatible versions

ktsitsi:
This PR:

- updates the logic of incremental ingestion in case of version conflict. The default behaviour in case of a version incompatibility does not error out rather than raise a warning and perform a no-op.
-


ihnorton (commented):
It seems like this check should be based on the format spec version rather than the package version.

ihnorton (approved):

PR #105, opened 2023-10-10
Logger for tiledb-bioimg

ktsitsi:
In this PR:

- We introduce a logger in `tiledb-bioimg` package for easier debugging. The logging in the current commit is mostly focused on OMETiffConverter, Reader, Writer and the corresponding APIs. This initial implementation bootstraps the logging of other Converters as well. The logging messages for the rest converters will be added in following commits.
- Due to the dual nature of our API (Wrappers and Converter functions) we introduce for the wrappers an extra argument called `verbose`. This boolean when set to `True`, will enable DEBUG mode across the function calls and when `False` it will just behave in WARNING level debugging. 
- For the Converters API (consisting of `to_tiledb` and `from_tiledb`) we introduce an extra arg called `log` as a `Union[bool, logging.Logger]`. The reasoning behind this is dual. First with a boolean arg we can use the calls standalone enabling their own logger . For the `logging.Logger` case we allow this way the integration with the wrapper API calls `from_bioimg` and `to_bioimg` as well as we enable the user to pass his/her own logger and integrate it with the rest of the package. 
- This PR will be connected to a PR in TileDB-Cloud-Py client so the logging could be enabled from the ingestion Graph as well.

A sample of the usage is below:
```
[2023-10-10 16:05:46,773] [helpers] [get_logger_wrapper] [DEBUG] tiledb=(0, 23, 0), libtiledb=(2, 17, 0), tiledb-bioimg=(0, 2, 4, 'dev12', 'dirty')
[2023-10-10 16:05:46,773] [base] [to_tiledb] [DEBUG] Max tiles:{'T': 1, 'C': inf, 'Z': 1, 'Y': 1024, 'X': 1024}
[2023-10-10 16:05:46,773] [base] [to_tiledb] [DEBUG] Updated max tiles:{'T': 1, 'C': inf, 'Z': 1, 'Y': 1024, 'X': 1024}
[2023-10-10 16:05:46,775] [base] [to_tiledb] [DEBUG] Format version: None
[2023-10-10 16:05:46,775] [ome_tiff] [level_count] [DEBUG] Level count: 2
[2023-10-10 16:05:46,775] [base] [to_tiledb] [DEBUG] Compressors : {0: ZstdFilter(level=0), 1: ZstdFilter(level=0)}
[2023-10-10 16:05:46,775] [base] [to_tiledb] [DEBUG] Convert arguments : {'reader': <tiledb.bioimg.converters.ome_tiff.OMETiffReader object at 0x1385b3430>, 'rw_group': <tiledb.bioimg.helpers.ReadWriteGroup object at 0x13868d7c0>, 'max_tiles': {'T': 1, 'C': inf, 'Z': 1, 'Y': 1024, 'X': 1024}, 'preserve_axes': False, 'chunked': False, 'max_workers': 0, 'compressor': {0: ZstdFilter(level=0), 1: ZstdFilter(level=0)}}
[2023-10-10 16:05:46,775] [ome_tiff] [image_metadata] [DEBUG] Image metadata: {'channels': [{'id': 'Channel:0:0', 'name': 'Channel 0', 'color': {'red': 255, 'green': 0, 'blue': 0, 'alpha': 255}}, {'id': 'Channel:0:1', 'name': 'Channel 1', 'color': {'red': 0, 'green': 255, 'blue': 0, 'alpha': 255}}, {'id': 'Channel:0:2', 'name': 'Channel 2', 'color': {'red': 0, 'green': 0, 'blue': 255, 'alpha': 255}}], 'physicalSizeX': 0.499, 'physicalSizeXUnit': 'µm', 'physicalSizeY': 0.499, 'physicalSizeYUnit': 'µm'}
[2023-10-10 16:05:46,775] [ome_tiff] [level_count] [DEBUG] Level count: 2
[2023-10-10 16:05:46,775] [base] [to_tiledb] [INFO] Converting level: 0
[2023-10-10 16:05:46,775] [ome_tiff] [axes] [DEBUG] Reader axes: Axes(dims='CYX')
[2023-10-10 16:05:46,776] [ome_tiff] [level_shape] [DEBUG] Level 0 shape: (3, 2967, 2220)
[2023-10-10 16:05:46,776] [ome_tiff] [axes] [DEBUG] Reader axes: Axes(dims='CYX')
[2023-10-10 16:05:46,776] [ome_tiff] [level_shape] [DEBUG] Level 0 shape: (3, 2967, 2220)
[2023-10-10 16:05:46,776] [ome_tiff] [level_dtype] [DEBUG] Level 0 dtype: uint8
```

### DIFF
--- a/tiledb/bioimg/helpers.py
+++ b/tiledb/bioimg/helpers.py
@@ -298,7 +298,7 @@ def get_logger_wrapper(
     :return: logger instance
     """
 
-    level = logging.DEBUG if verbose else logging.INFO
+    level = logging.DEBUG if verbose else logging.WARNING
     logger = get_logger(level)
 
     logger.debug(
--- a/tiledb/bioimg/helpers.py
+++ b/tiledb/bioimg/helpers.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 import os
 import sys
 from pathlib import Path
@@ -14,6 +15,7 @@ from tiledb.cc import WebpInputFormat
 
 from . import ATTR_NAME
 from .converters.axes import Axes, AxesMapper
+from .version import version_tuple
 
 
 class ReadWriteGroup:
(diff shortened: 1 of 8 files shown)


tiledb/bioimg/helpers.py
```diff
@@ -260,3 +262,50 @@ def is_win_path(scheme: str) -> bool:
...
+    verbose: bool = False,
+) -> logging.Logger:
+    """
+    Get a logger instance and log version information.
+
+    :param verbose: verbose logging, defaults to False
+    :return: logger instance
+    """
+
+    level = logging.DEBUG if verbose else logging.INFO
```
ihnorton, line 301: Am I following correctly that this will _always_ be set to `logging.INFO`? I think we need to be able to turn it off entirely (by default)...

ktsitsi: The default is indeed `logging.info` level. Even if the info logs are minimal I am totally positive to modify the default logger lever to `warning` so it turns off entirely. What do you think?

ihnorton: I would lean toward disabling by default unless you feel strongly otherwise - will defer to you.

PR #113, opened 2024-03-04
Tiled writes during ingestion

XanthosXanthopoulos:
This PR adds tiled writes to TileDB arrays during non chunked ingestion. This change allows non chunked ingestion to run directly against a TileDB uri, fixing the `Entity too Large` constraint in REST. To avoid small writes and achieve network saturation the write domain tile can be scaled to write larger amounts of data at once.


ktsitsi (approved):
LGTM just a missed typo, and fix CI.

ktsitsi (approved):

XanthosXanthopoulos:
Write TileDB array in chunks [sc-42410]

PR #115, opened 2024-03-05
Fix duplicate channel ID

XanthosXanthopoulos:
This PR fixes a bug where a duplicate channel ID will be written in case of an image a channel with multiple samples per pixel


ktsitsi (approved):
LGTM

PR #120, opened 2024-05-02
Memory layout optimized TIFF reader

XanthosXanthopoulos:
TIFF files are either stored in striped or tiled format. Each segment (either tile or strip) is stored independently of the other segments which means that the logical order of the segments does not match the physical ordering inside the file. 

When we do chunked ingestion the image chunk we request is not contiguous in memory and the tifffile library does one request per segment to construct the image tile we request. For files stored on s3 the access pattern in suboptimal (many very small requests generally around 4KB).

This PR attempts to fix this by introducing a caching mechanism. The image file is read sequentially and each segment(compressed) is stored into a cache. Once enough segments are retrieved such that they fill an entire dimension, they are decoded and written to TileDB. The image reading and writing are decoupled but communicating the cap the memory usage and use resources more optimally.


ktsitsi (changes requested):
Generally LGTM just some minor things.

ktsitsi (approved):

PR #123, opened 2024-05-30
Pushing VFS down to converters

ktsitsi:
This PR:

- Pushes down the VFS inside the converters allowing passing of config and context for accessing input images directly from S3. 

Usage:
```bash
from tiledb.bioimg import from_bioimg
from_bioimg(src='s3://tiledb-bioimg/local_testing/CMU-1-Small-Region-rgb.ome.tiff', dest='./tests/data/cmu-rgb-vfs.tdb', reader_kwargs={"config":tiledb.Config({})}
>>Ingesting level 0: 100%|██████████| 9/9 [00:00<00:00, 47.70tiles/s]
>>Ingesting level 1: 100%|██████████| 1/1 [00:00<00:00, 47.95tiles/s]

---
from tiledb.bioimg.openslide import TileDBOpenSlide
slide = TileDBOpenSlide('s3://tiledb-bioimg/local_testing/CMU-1-Small-Region-rgb.ome.tiff', config=tiledb.Config({}))
slide.levels
>>Out[9]: (0, 1, 2, 3)

```

It will resolve related story [sc-25684] [sc-43577]

### DIFF
--- a/tiledb/bioimg/converters/ome_tiff.py
+++ b/tiledb/bioimg/converters/ome_tiff.py
@@ -8,7 +8,7 @@ import numpy as np
 import tifffile
 
 from tiledb.cc import WebpInputFormat
-
+from tiledb import VFS, Config, Ctx
 from .. import ATTR_NAME, EXPORT_TILE_SIZE, WHITE_RGBA
 from ..helpers import get_decimal_from_rgba, get_logger_wrapper, get_rgba, iter_color
 from .axes import Axes
@@ -21,6 +21,8 @@ class OMETiffReader(ImageReader):
         self,
         input_path: str,
         logger: Optional[logging.Logger] = None,
+        config: Optional[Config] = None,
+        ctx: Optional[Ctx] = None,
         extra_tags: Sequence[Union[str, int]] = (),
     ):
         """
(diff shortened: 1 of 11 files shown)


tiledb/bioimg/converters/openslide.py
```diff
@@ -13,23 +13,39 @@
...
         """
+        self._ctx = _get_ctx(ctx, config)
+        self._cfg = self._ctx.config()
         self._logger = get_logger_wrapper(False) if not logger else logger
-        self._osd = osd.OpenSlide(input_path)
+        resolved_path = input_path
+        if is_remote_protocol(input_path):
+            resolved_path = cache_filepath(
+                input_path, config, ctx, self._logger, scratch_space
+            )
```
jp-dark, line 47:
```suggestion
        if is_remote_protocol(input_path):
            resolved_path = cache_filepath(
                input_path, config, ctx, self._logger, scratch_space
            )
        else:
            resolved_path = input_path
```

tiledb/bioimg/converters/ome_tiff.py
```diff
@@ -21,6 +22,8 @@ def __init__(
         self,
         input_path: str,
         logger: Optional[logging.Logger] = None,
+        config: Optional[Config] = None,
+        ctx: Optional[Ctx] = None,
         extra_tags: Sequence[Union[str, int]] = (),
     ):
```
jp-dark, line 28:
```suggestion
        self,
        input_path: str,
        *,
        logger: Optional[logging.Logger] = None,
        config: Optional[Config] = None,
        ctx: Optional[Ctx] = None,
        extra_tags: Sequence[Union[str, int]] = (),
    ):
```
It might be worth making optional arguments keyword only so that you don't need to worry about breaking the API when adding in new optional arguments. As implemented, if someone naively tried used the old API with only positional arguments, it would just pass `extra_tags` to `config` without a clear error message.

jp-dark (approved):
Looks good to me (left one optional comment, but feel free to implement or ignore).

PR #125, opened 2024-06-12
Fixing API dependencies guards and tiledb-cloud warning

ktsitsi:
This PR:

- Fixes optional dependencies introducing guards in tiledb-bioimg API.
- Removes `tiledb-cloud` from the full installation. This will not affect our deployments in docker images or notebooks since the `tiledb-cloud-py` is already installed there. This will remove the warning - even though we were not importing the cloud-client anywhere- when a user is not logged in.
- Adding some warnings in case optional dependencies are missing.
- With all optional dependencies removed the user will face some warning but the `tiledb.bioimg` will be imported.
- Only when the user access the a Converter that relies on a dependency not present he gets a `ModuleNotFoundError`

```bash
In [1]: import tiledb.bioimg
/Users/konstantinostsitsimpikos/tiledb_dev/TileDB-BioImaging/./tiledb/bioimg/converters/ome_tiff.py:12: UserWarning: OMETiff Converter requires 'tifffile' package. You can install 'tiledb-bioimg' with the 'tiff' or 'full' flag
  warnings.warn(
/Users/konstantinostsitsimpikos/tiledb_dev/TileDB-BioImaging/./tiledb/bioimg/converters/ome_zarr.py:17: UserWarning: OMEZarr Converter requires 'ome-zarr' package. You can install 'tiledb-bioimg' with the 'zarr' or 'full' flag
  warnings.warn(
/Users/konstantinostsitsimpikos/tiledb_dev/TileDB-BioImaging/./tiledb/bioimg/converters/openslide.py:25: UserWarning: Openslide Converter requires 'openslide-python' package. You can install 'tiledb-bioimg' with the 'openslide' or 'full' flag
  warnings.warn(
In [2]: import tiledb.bioimg
In [3]: from tiledb.bioimg import from_bioimg
```

### DIFF
--- a/tiledb/bioimg/converters/ome_zarr.py
+++ b/tiledb/bioimg/converters/ome_zarr.py
@@ -10,7 +10,7 @@ import numpy as np
 
 try:
     import zarr
-    from numcodecs import Blosc
+    from zarr.codecs import Blosc
     from ome_zarr.reader import OMERO, Multiscales, Reader, ZarrLocation
     from ome_zarr.writer import write_multiscale
 except ImportError as err:
--- a/tiledb/bioimg/converters/ome_zarr.py
+++ b/tiledb/bioimg/converters/ome_zarr.py
@@ -10,9 +10,9 @@ import numpy as np
 
 try:
     import zarr
-    from zarr.codecs import Blosc
     from ome_zarr.reader import OMERO, Multiscales, Reader, ZarrLocation
     from ome_zarr.writer import write_multiscale
+    from zarr.codecs import Blosc
 except ImportError as err:
     warnings.warn(
         "OMEZarr Converter requires 'ome-zarr' package. "
--- a/tiledb/bioimg/converters/ome_zarr.py
+++ b/tiledb/bioimg/converters/ome_zarr.py
@@ -2,14 +2,23 @@ from __future__ import annotations
 
 import json
 import logging
+import warnings
 from typing import Any, Dict, List, Mapping, Optional, Sequence, Tuple, cast
 
 import numpy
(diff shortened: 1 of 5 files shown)


tiledb/bioimg/converters/ome_zarr.py
```diff
@@ -2,14 +2,23 @@
...
 import numpy
 import numpy as np
-import zarr
-from numcodecs import Blosc
-from ome_zarr.reader import OMERO, Multiscales, Reader, ZarrLocation
-from ome_zarr.writer import write_multiscale
+
+try:
+    import zarr
+    from numcodecs import Blosc
```
jp-dark, line 13: Should the numcodecs import be outside of this try block? If it fails it will warn about the wrong module.

jp-dark (approved):
I have one comment on a zarr import error, but other than that looks good to me.

tiledb/bioimg/converters/ome_zarr.py
ktsitsi: Context-free that is true. However the `numcodecs` is a direct dependency of `zarr` and installing the latter installs the former. Thank you for this comment though cause based on this one I will change this import from `from numcodecs import Blosc` to `from zarr.codecs import Blosc` where they just wrap the `numcodecs.Blosc` importation. This way it will be cleaner.

```python

# site-packages/zarr/codecs.py
# flake8: noqa
from numcodecs import *
from numcodecs import get_codec, Blosc, Pickle, Zlib, Zstd, Delta, AsType, BZ2
from numcodecs.registry import codec_registry
```

PR #128, opened 2024-06-18
Pinning scikit-image in Bioimg CI

ktsitsi:


ihnorton (approved):
Let's file a follow-up story to check and remove this pin in a week or two.

PR #129, opened 2024-07-02
Validating ingestion of levels to make ingestion idempotent

ktsitsi:
This PR:

- Introduces a helper function `validate_ingestion` that compares the `ned` and the number of fragments of each level with the `array schema` domains of each dimension to impose a correctness check.
- Till now once a level was created we were assuming its correctness without any check in order to support `incremental ingestion`. This logic however contradicted with the retry policy imposed during batch ingestion in taskgraph environment and led to missing metadata and errors like the following:

```python
 File "tiledb/libmetadata.pyx", line 327, in tiledb.libtiledb.Metadata.__getitem__
KeyError: 'level'
```
-  Tests the validation function's internals in unittest.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,11 +1,13 @@
 from pathlib import Path
 
+import random
 import numpy as np
 from skimage.metrics import structural_similarity
 
 import tiledb
 from tiledb.bioimg import ATTR_NAME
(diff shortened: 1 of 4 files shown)


johnkerl (approved):
🚢

PR #130, opened 2024-07-02
Remove pinned scikit since 0.24.0 was released

ktsitsi:
This PR:

- Removes the pin in CI for scikit-image after the release of version [`0.24.0`](https://pypi.org/project/scikit-image/0.24.0/) that supports `numpy==2.0`


jp-dark (approved):

ihnorton:
Thanks!

PR #133, opened 2024-07-16
Convert ABC base classes to Protocols

ktsitsi:
This PR:

- Modifies the `ImageReader` and `ImageWriter` from using abstract base classes to use protocols. This will allow third-party plugins to define `ImageReader` and `ImageWriter` classes without importing `TileDB-BioImaging`.

### DIFF
--- a/tests/integration/converters/test_ome_zarr.py
+++ b/tests/integration/converters/test_ome_zarr.py
@@ -8,8 +8,7 @@ import zarr
 import tiledb
 from tests import assert_image_similarity, get_path, get_schema
 from tiledb.bioimg.converters import DATASET_TYPE, FMT_VERSION
-from tiledb.bioimg.converters.ome_tiff import OMETiffReader
-from tiledb.bioimg.converters.ome_zarr import OMEZarrConverter, OMEZarrReader
+from tiledb.bioimg.converters.ome_zarr import OMEZarrConverter
 from tiledb.bioimg.helpers import iter_color, open_bioimg
 from tiledb.bioimg.openslide import TileDBOpenSlide
 from tiledb.cc import WebpInputFormat
@@ -25,11 +24,11 @@ def test_ome_zarr_converter_source_reader_exception(
     tiff_path = get_path("CMU-1-Small-Region.ome.tiff")
     output_reader = tmp_path / "to_tiledb_reader"
 
-    with pytest.raises(ValueError) as excinfo:
+    with pytest.raises(FileExistsError) as excinfo:
         OMEZarrConverter.to_tiledb(
-            OMETiffReader(tiff_path), str(output_reader), preserve_axes=preserve_axes
+            tiff_path, str(output_reader), preserve_axes=preserve_axes
         )
-    assert "reader should match converter" in str(excinfo)
+    assert "FileExistsError" in str(excinfo)
 
 
 @pytest.mark.parametrize("series_idx", [0, 1, 2])
@@ -46,7 +45,7 @@ def test_ome_zarr_converter_reader_source_consistent_output(
         input_path, str(output_path), preserve_axes=preserve_axes
     )
     OMEZarrConverter.to_tiledb(
-        OMEZarrReader(input_path), str(output_reader), preserve_axes=preserve_axes
+        input_path, str(output_reader), preserve_axes=preserve_axes
     )
 
     # check the first (highest) resolution layer only
@@ -62,7 +61,7 @@ def test_ome_zarr_converter_reader_source_consistent_output(
 @pytest.mark.parametrize("series_idx", [0, 1, 2])
 @pytest.mark.parametrize("preserve_axes", [False, True])
 def test_ome_zarr_converter(tmp_path, series_idx, preserve_axes):
-    input_path = get_path("CMU-1-Small-Region.ome.zarr") / str(series_idx)
+    input_path = get_path("CMU-1-Small-Region.ome" ".zarr") / str(series_idx)
     OMEZarrConverter.to_tiledb(input_path, str(tmp_path), preserve_axes=preserve_axes)
 
     # check the first (highest) resolution layer only
--- a/tests/integration/converters/test_ome_zarr.py
+++ b/tests/integration/converters/test_ome_zarr.py
@@ -61,7 +61,7 @@ def test_ome_zarr_converter_reader_source_consistent_output(
 @pytest.mark.parametrize("series_idx", [0, 1, 2])
 @pytest.mark.parametrize("preserve_axes", [False, True])
 def test_ome_zarr_converter(tmp_path, series_idx, preserve_axes):
-    input_path = get_path("CMU-1-Small-Region.ome" ".zarr") / str(series_idx)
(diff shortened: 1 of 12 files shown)


jp-dark (changes requested):
I think you need to remove the `__init__` methods and push down anything done their to the subclasses. For the ImageWriter, you might find creating a Logger mixin useful.

ktsitsi:
I would really appreciate a review pass with focus especially on the typing side. Another commit will follow with the plugin support using entrypoints.

jp-dark (approved):
This looks good to me, although I'm not super familiar with this code base so I may be missing something. The protocols should be workable on the TileDB-SOMA side.

ktsitsi:
@thetorpedodog I would really appreciate some of your time here mostly on typing.

tiledb/bioimg/converters/base.py
```diff
@@ -224,13 +204,28 @@ def write_level_image(
...
+
 
+class ImageConverter(Protocol[TReader, TWriter]):
+    _ImageReaderType: Optional[Type[TReader]] = None
+    _ImageWriterType: Optional[Type[TWriter]] = None
 
-class ImageConverter:
+    def from_tiledb(self) -> ImageConverter[TReader, TWriter]: ...
+
+    def to_tiledb(self) -> ImageConverter[TReader, TWriter]: ...
```
thetorpedodog, line 221: I don’t quite follow what’s going on here. An ImageConverter has a method that returns another thing of its own type? This annotation would indicate to me that I could do this:

```python
converter = build_some_converter()
other_converter = converter.from_tiledb()
third_converter = converter.to_tiledb()
# and even
more_converter = third_converter.from_tiledb()
```

the type of all of which would be some form of `ImageConverter[...]`.

Based on the below, it looks like you might be intending to make these `@classmethod`s with at least a subset of the parameters below.

tests/integration/converters/test_ome_zarr.py
```diff
@@ -62,7 +61,7 @@ def test_ome_zarr_converter_reader_source_consistent_output(
 @pytest.mark.parametrize("series_idx", [0, 1, 2])
 @pytest.mark.parametrize("preserve_axes", [False, True])
 def test_ome_zarr_converter(tmp_path, series_idx, preserve_axes):
-    input_path = get_path("CMU-1-Small-Region.ome.zarr") / str(series_idx)
+    input_path = get_path("CMU-1-Small-Region.ome" ".zarr") / str(series_idx)
```
thetorpedodog, line 64: an extra `" "` somehow ended up in here

tiledb/bioimg/converters/base.py
```diff
@@ -58,80 +60,64 @@
...
-        dest_config: Optional[tiledb.Config] = None,
-        dest_ctx: Optional[tiledb.Config] = None,
-        **kwargs: Any,
-    ):
-        """Initialize this ImageReader"""
+class ImageReader(Protocol):
 
-    def __enter__(self) -> ImageReader:
-        return self
+    def __enter__(self) -> ImageReader: ...
```
thetorpedodog, line 65: I imagine you’ll want to make this a `Self`, since that explicitly says “the thing you return here is of the same type as me”. Right now it would be allowable (if strange) to have

```python
reader = ImageReaderA()
with reader as with_block_reader:
  reveal_type(with_block_reader)  # ImageReaderB
```

tiledb/bioimg/converters/base.py
```diff
@@ -58,80 +60,64 @@
...
-    @abstractmethod
     def logger(self) -> Optional[logging.Logger]:
         """The logger of this image reader."""
+        ...
 
     @logger.setter
-    @abstractmethod
     def logger(self, default_logger: logging.Logger) -> None:
         """The setter for the logger of this image reader"""
+        ...
```
thetorpedodog, line 87: Since this is gettable/settable, you could [make this a simple property](https://typing.readthedocs.io/en/latest/spec/protocol.html#protocols:~:text=class%20Template(Protocol)%3A%0A%20%20%20%20name%3A%20str%20%20%20%20%20%20%20%20%23%20This%20is%20a%20protocol%20member%0A%20%20%20%20value%3A%20int%20%3D%200%20%20%20%23%20This%20one%20too%20(with%20default)):

```python
class ImageReader(Protocol):
  ...
  logger: Optional[logging.Logger]
  ...
```

tiledb/bioimg/converters/base.py
```diff
@@ -288,7 +283,9 @@ def from_tiledb(
             output_config = config
 
         slide = TileDBOpenSlide(input_path, attr=attr, config=config)
-        writer = cls._ImageWriterType(destination_uri, logger)
+        writer = cls._ImageWriterType(
+            destination_uri, logger, **writer_kwargs if writer_kwargs else {}
```
thetorpedodog, line 287: `**(writer_kwargs or {})` might be a little clearer here.

tiledb/bioimg/plugin_manager.py
```diff
@@ -0,0 +1,28 @@
...
+        for entry_point in importlib.metadata.entry_points()["bioimg.writers"]:
+            writers[entry_point.name] = entry_point.load()
+        return writers
+
+    @staticmethod
+    def load_converters() -> Mapping[str, Type[ImageConverterMixin[Any, Any]]]:
+        converters = {}
+        for entry_point in importlib.metadata.entry_points()["bioimg.converters"]:
+            converters[entry_point.name] = entry_point.load()
+        return converters
```
thetorpedodog, line 28: Since these are all static methods, they don’t really need to live inside a class; they can just be module-level functions.

You could also save some repetition with something like

```python
def load_readers() -> ...:
  return _load_entrypoints("bioimg.readers")


def _load_entrypoints(name: str) -> Mapping[str, Any]:
  eps = importlib.metadata.entry_points()[name]
  return {ep.name: ep.load() for ep in eps}
```

tiledb/bioimg/wrappers.py
```diff
@@ -1,29 +1,29 @@
 from __future__ import annotations
 
-from typing import Any, Type
-
-from .converters.base import ImageConverter
+from importlib import util
```
thetorpedodog, line 3: might make sense to have this be `import importlib.util` so that later, it’s easier to understand the exact `util` you’re referring to.

thetorpedodog (commented):
Very cool! I am a fan of the use of `entry_points` to make the converters completely pluggable. The general structure is pretty good; I have some feedback about specifics.

tiledb/bioimg/converters/base.py
ktsitsi: Sure I will use the `TReader` and `TWriter` respectively as the supported version is `pre-3.11`

thetorpedodog:
I somehow managed to leave comments in the wrong place (apparently you can just comment on a changeset floating in space?), but you can read them here: https://github.com/TileDB-Inc/TileDB-BioImaging/commit/03217bf7d32ab49d6a17b6234c193c50dd660236

the biggest thing: you can use `Self` from `typing_extensions` and that will make a lot of other stuff less verbose.

PR #134, opened 2024-07-17
Support source dest configs ingestion

ktsitsi:
This PR:

- Adds support for separate source `config` and destination `config` for the ingestion.

**Depends on `tiledb-py==0.30.3` release**

### DIFF
--- a/tiledb/bioimg/converters/base.py
+++ b/tiledb/bioimg/converters/base.py
@@ -319,7 +319,7 @@ class ImageConverter:
         exclude_metadata: bool = False,
         compressor: Optional[Union[Mapping[int, Any], Any]] = None,
         log: Optional[Union[bool, logging.Logger]] = None,
-        reader_kwargs: Optional[Mapping[str, Any]] = None,
+        reader_kwargs: Optional[MutableMapping[str, Any]] = None,
         pyramid_kwargs: Optional[Mapping[str, Any]] = None,
     ) -> Type[ImageConverter]:
         """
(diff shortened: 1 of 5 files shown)


tiledb/bioimg/converters/base.py
```diff
@@ -62,8 +62,10 @@ def __init__(
         input_path: str,
         *,
         logger: Optional[logging.Logger],
-        config: Optional[tiledb.Config] = None,
-        ctx: Optional[tiledb.Ctx] = None,
+        source_config: Optional[tiledb.Config] = None,
```
ihnorton, line 65: Is this a breaking change of the current external API surface? If so, we should avoid that by providing backward-compatibility for `config (it will complicate the code a little bit, but we can make a single handler to process all three kwargs)

ktsitsi: Indeed it is a breaking change for the `to_tiledb` API . I tried to keep the `from_bioimg` API backwards compatible where I did as you recommend here. I guess the `to_tiledb` belongs to the external API too right? I can propagate it down there. Since we have not yet provided any backwards compatibility guarantees for these function I thought of ditching it in favour of readability but I can make the changes.

ihnorton: Yes. I think we should have a fallback such that `config` sets both `read_config` and `write_config` if they are not supplied.

ktsitsi: I would keep two with the writer_config being the same with the reader_config if the former is not supplied. With None being the default. How about this logic? It is similar to what I ve done for the wrappers API.

PR #136, opened 2024-09-06
Support ingestion of PNG images

ktsitsi:
This PR:

- Adds support for the ingestion of PNGs
- Creates `Reader`, `Writer` and `Converter` for PNG images using the `PIL` package.
- Updates the upper level API `from_bioimg` to use the `PNGConverter`
- Unit tests / Integration tests / Roundtrip ingestion-exportation


tests/integration/converters/test_png.py
```diff
@@ -0,0 +1,111 @@
...
+def test_png_converter(tmp_path):
+    input_path = str(get_path("A01_s1--cell_outlines.png"))
+    output_path = str(tmp_path)
+
+    PNGConverter.to_tiledb(input_path, output_path)
+
+    with TileDBOpenSlide(output_path) as t:
+        assert len(tiledb.Group(output_path)) == t.level_count == 1
+        schemas = get_schema(1080, 1080)
+        assert t.dimensions == schemas.shape[:-3:-1]
```
johnkerl, line 25: The magic numbers -3 and -1 here could use an explanatory comment

tests/integration/converters/test_png.py
```diff
@@ -0,0 +1,111 @@
...
+        for level in range(t.level_count):
+            region_data = t.read_region((0, 0), level, t.level_dimensions[level])
+            level_data = t.read_level(level)
+            np.testing.assert_array_equal(region_data, level_data)
+
+
+@pytest.mark.parametrize("filename", ["A01_s1--cell_outlines.png"])
+def test_png_converter_group_metadata(tmp_path, filename):
+    input_path = get_path(filename)
+    tiledb_path = tmp_path / "to_tiledb"
```
johnkerl, line 47:
```suggestion
    tiledb_path = str(tmp_path / "to_tiledb")
```

and then you won't need the two `str(tiledb_path)` below

tests/integration/converters/test_png.py
```diff
@@ -0,0 +1,111 @@
...
+    PNGConverter.to_tiledb(input_path, str(tiledb_path), preserve_axes=False)
+
+    with TileDBOpenSlide(str(tiledb_path)) as t:
+        group_properties = t.properties
```
johnkerl, line 51: Just curious, any reasaon not to call this `t.metadata`? I understand this may be an already-committed-to design decision, which is fine -- I'm not asking you to change anything -- just curious if "properties" is a name chosen to align with some UX which is familiar to the target scientific audience ...

tests/integration/converters/test_png.py
```diff
@@ -0,0 +1,111 @@
...
+        assert group_properties["dataset_type"] == DATASET_TYPE
+        assert group_properties["fmt_version"] == FMT_VERSION
+        assert isinstance(group_properties.get("pkg_version"), str)
+        assert group_properties["axes"] == "XYC"
```
johnkerl, line 55: also curious why this one and only this one is done using `.get()` rather than `[...]` ...

tests/integration/converters/test_png.py
```diff
@@ -0,0 +1,111 @@
...
+        (tiledb.WebpFilter(WebpInputFormat.WEBP_RGB, lossless=False), False),
+        (tiledb.WebpFilter(WebpInputFormat.WEBP_RGB, lossless=True), True),
+        (tiledb.WebpFilter(WebpInputFormat.WEBP_NONE, lossless=True), True),
+    ],
+)
+def test_png_converter_roundtrip(
+    tmp_path, filename, preserve_axes, chunked, compressor, lossless
+):
+    input_path = get_path(filename)
+    tiledb_path = tmp_path / "to_tiledb"
```
johnkerl, line 99:
```suggestion
    tiledb_path = str(tmp_path / "to_tiledb")
```

tests/integration/converters/test_png.py
```diff
@@ -0,0 +1,111 @@
...
+    output_path = tmp_path / "from_tiledb"
```
johnkerl, line 100:
```suggestion
    output_path = str(tmp_path / "from_tiledb")
```

tiledb/bioimg/converters/png.py
```diff
@@ -0,0 +1,215 @@
...
+        return l_shape
+
+    @property
+    def webp_format(self) -> WebpInputFormat:
+        self._logger.debug(f"Keyframe photometric: {self._png.mode}")
+        if self._png.mode == "RGB":
+            return WebpInputFormat.WEBP_RGB
+        return WebpInputFormat.WEBP_NONE
+
+    def level_image(
```
johnkerl, line 123: Unless this naming is already agreed-on / committed to, I'd suggest changing it.

In English this is ambiguous, since "level" is a noun, an adjective, and a verb. The name might mean:

* Is the image something that is level? Apparently not since the return type isn't `bool`
* Return a copy of a non-level image, for which the copy is level
* Make the image level in-place -- apparently not given the return type
* Extract one level out of a multi-image level

I'd suggest maybe calling `get_image_level`, but now I'm confused since the `level` argument of the method isn't used within the method body ...

Also, please add a docstring

tiledb/bioimg/converters/png.py
```diff
@@ -0,0 +1,215 @@
...
+        self, level: int = 0, tile: Optional[Tuple[slice, ...]] = None
+    ) -> np.ndarray:
+
+        if tile is None:
+            return np.asarray(self._png)
+        else:
+            return np.asarray(self._png)[tile]
+
+    def level_metadata(self, level: int) -> Dict[str, Any]:
```
johnkerl, line 132: Same two concerns as `level_image`

tiledb/bioimg/converters/png.py
```diff
@@ -0,0 +1,215 @@
...
+                "id": f"{idx}",
+                "name": f"Channel {idx}",
+                "color": (next(color_generator)),
+            }
+            channels.append(channel_metadata)
+        self._metadata["channels"] = channels
+        return self._metadata
+
+    @property
+    def original_metadata(self) -> Dict[str, Any]:
```
johnkerl, line 160: It ssems like `group_metadata` is a generalization of this ...

* Maybe put the two next to each other
* Maybe have one call the other (no big deal, they're already short functions)
* Please add docstrings

johnkerl (changes requested):
Comments interleaved.

I don't know this package well but as a newcomer I would suggest more use of method docstrings ...

tests/integration/converters/test_png.py
ktsitsi: As you assumed the names chosen for all the functions/properties align with the [Openslide API](https://openslide.org/api/python/#openslide.OpenSlide).  This applies to `level_metadata`, `level_image`, `properties` and pretty much any other function name.

tiledb/bioimg/converters/png.py
ktsitsi: https://github.com/TileDB-Inc/TileDB-BioImaging/pull/136#discussion_r1753840251

tiledb/bioimg/converters/png.py
ktsitsi: It happens in this specific Converter that the two coincide. It is indeed repeated information but we need to follow the design as it is. PNGs are not pyramidal whole slides so incorporating the converter into a design that was mostly built for multi-resolution imagery requires that. I can modify so one function calls the other.

tests/integration/converters/test_png.py
ktsitsi: They are equivalent I guess it is a leftover of different contributors. Mostly the tests we perform for all the converters are identical so some copy-paste there. I can change so it is consistent.

tiledb/bioimg/converters/png.py
ktsitsi: On doctstrings comment:  All these `converters/ readers/ writers` follow the base protocols `ImageReader`, `ImageWriter`, `ImageConverter` in `base.py` . This way they implement the protocol and the `docstrings` for each of these functions/properties resides there to avoid repetition and it is common for all implementations.

johnkerl (approved):
LGTM except missing docstrings, and if that's how this package is, with or without this PR, that's nothing wrong with this PR. (Per Slack it sounds like the docstrings are on the base classes, not the derived classes.)

tiledb/bioimg/converters/png.py
```diff
@@ -109,7 +109,10 @@ def level_shape(self, level: int = 0) -> Union[Tuple[int, ...], Any]:
                 "PNG images are not pyramidal and consist of only one level"
             )
         # Even after converting to RGB the size is not updated to 3d from 2d
-        l_shape = (*self._png.size, 3)
+        w, h = self._png.size
+
+        # Numpy shape is of the format (H, W, D) compared to Pillow (W, H, D)
+        l_shape = (h, w, 3)
```
jp-dark, line 115: I think this might still cause problems. I believe png supports 1-4 channels depending on the file. CZI definitely was importing some RGBA data.

ktsitsi: If transparency is relevant and needed I can definitely modify the code to support it.

jp-dark: I think it will be necessary on the SOMA side.

XanthosXanthopoulos (dismissed):

tiledb/bioimg/converters/png.py
```diff
@@ -0,0 +1,215 @@
...
+        self._png = Image.open(self._vfs_fh)
+
+        # Get initial metadata
+        self._metadata: Dict[str, Any] = self._png.info
+        self._metadata.update(dict(self._png.getexif()))
+
+        # Handle all different modes as RGB for consistency
+        if self._png.mode not in ["RGB", "RGBA"]:
+            self._metadata["original_mode"] = self._png.mode
+            self._png = self._png.convert("RGB")
```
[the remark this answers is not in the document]
ktsitsi: The same will apply for `PA` right?

jp-dark (approved):
This worked for all the single cell data I tried it on (including one with alpha channels).

PR #137, opened 2024-09-18
Exported images fail to render in qupath

ktsitsi:
Problem statement:
You can find it in the shortcut ticket.

This PR:

- Parses the OME-XML metadata and instead of removing it as a whole it removes only the metadata for the `macro` and `label` images. Leaving the OME-XML metadata structure solves the issue with Qupath and the latter can select the correct reader for rendering an exported image.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -8,6 +8,7 @@ import tiledb
 from tiledb.bioimg import ATTR_NAME
 from tiledb.cc import WebpInputFormat
 from tiledb.bioimg.helpers import merge_ned_ranges
+import xml.etree.ElementTree as ET
 
 DATA_DIR = Path(__file__).parent / "data"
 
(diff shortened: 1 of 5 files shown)


tests/__init__.py
```diff
@@ -111,3 +112,96 @@ def generate_test_case(num_axes, num_ranges, max_value):
...
+
+    # Create the root element
+    ome = ET.Element(
+        root_tag,
+        {
+            "xmlns": "http://www.openmicroscopy.org/Schemas/OME/2016-06",
+            "xmlns:xsi": "http://www.w3.org/2001/XMLSchema-instance",
+            "Creator": "tifffile.py 2023.7.4",
+            "UUID": "urn:uuid:40348664-c1f8-11ee-a19b-58112295faaf",
+            "xsi:schemaLocation": "http://www.openmicroscopy.org/Schemas/OME/2016-06 http://www.openmicroscopy.org/Schemas/OME/2016-06/ome.xsd",
```
spencerseale, line 128: Are these liable to change? Would we want to pop them or other specification out into a toml config or something similar?

tiledb/bioimg/helpers.py
```diff
@@ -456,3 +458,31 @@ def merge_ned_ranges(
     merged_ranges_per_axis = [merge_ranges(ranges) for ranges in ranges_per_axis]
 
     return tuple(merged_ranges_per_axis)
+
+
+def remove_ome_image_metadata(xml_string: str) -> Union[str, Any]:
```
spencerseale, line 463: doc strings if this needs to be rendered in an api ref eventually

spencerseale (approved):

tests/__init__.py
ktsitsi: I fixed it using some other `asserts` making it idempotent so now versions and headers should not make difference. Thanks!

tiledb/bioimg/helpers.py
ktsitsi: Added docstrings for the `remove_ome_image_metadata ` function although it is an internal function. Nice habits.!

PR #138, opened 2024-09-26
Add Prereq to install OpenSlide to README

spencerseale:


spencerseale:
Is CI job fail okay to proceed with merge?

ktsitsi:
> Is CI job fail okay to proceed with merge?

It was not your fault. Something is wrong or changed in the configuration of the environment that the action creates only on Windows runner.  I pinned a downgraded version and works now I opened also an issue here to track it. https://github.com/mamba-org/setup-micromamba/issues/227

PR #140, opened 2024-10-23
Adding a metadata flag for signaling successful ingestion

ktsitsi:
This PR:

- Adds a metadata KV at a group level of an ingested image signalling if the ingestion was completed successfully after all the steps.
- The metadata key `valid` gets a boolean value of either `True` or `False`
- During every ingestion a check is being performed to find if the desination path already exists and is a group. In this scenario we check if the group's metadata contain the key `valid`. If not this signals a corrupted ingested file (possibly from a previous run) and the response to that is for the group to be overwritten from scratch.
- In case the group exists but the ingestion runs in order for more layers to be added into the image `incremental ingestion` the flag is not enabled. We initially remove again the `valid` KV from the group's metadata until re-instantiate at the end of the current ingestion that is being running.

### DIFF
--- a/tiledb/bioimg/converters/base.py
+++ b/tiledb/bioimg/converters/base.py
@@ -410,6 +410,22 @@ class ImageConverterMixin(Generic[TReader, TWriter]):
         metadata = {}
         original_metadata = {}
 
+        # Checks group validity
+        overwrite = False
+        with rw_group:
+            if not rw_group.valid_group:
+                logger.debug(
+                    f"Already existing on disk Group at {output_path} evaluated as corrupted."
+                )
+                # Overwrite destination group
+                rw_group.m_group.delete(recursive=True)
+                overwrite = True
+
+        if overwrite:
+            # Re-initializes the destination Group for re-ingestion
+            logger.debug(f"Group {output_path} will be overwritten.")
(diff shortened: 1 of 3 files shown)


XanthosXanthopoulos (approved):

tiledb/bioimg/converters/base.py
```diff
@@ -410,6 +410,26 @@ def to_tiledb(
         metadata = {}
         original_metadata = {}
 
+        # Checks group validity
+        overwrite = False
```
spencerseale, line 414: Should `overwrite` be an arg that defaults to `False`? That would give us flexibility to overwrite every image. Although uncommon, may be times when we need to force overwrite. Could then make that arg accessible from `tiledb.cloud.bioimage.ingest` s well.

spencerseale: The caveat would be of course that if an image is invalid it will still be overwritten even if `overwrite=False` set in the caller

spencerseale (approved):

tiledb/bioimg/converters/base.py
ktsitsi: Yeap the direction of this patch is to mitigate retrial issues and corrupted leftovers, that the user fails to recognize. If users really want to overwrite/re-ingest an image on purpose they can always do it by changing the destination path and manually deleting the previous record.

PR #151, opened 2025-02-12
Extra pruning of Structured Annotations

ktsitsi:
This PR:

- Supports pruning for CommentAnnotations in the OME-XML metadata before ingestion.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -115,7 +115,9 @@ def generate_test_case(num_axes, num_ranges, max_value):
     return input_ranges, expected_output
 
 
-def generate_xml(has_macro=True, has_label=True, root_tag="OME", num_images=1):
+def generate_xml(
+    has_macro=True, has_label=True, has_annotations=True, root_tag="OME", num_images=1
+):
     """Generate synthetic XML strings with options to include 'macro' and 'label' images."""
 
     # Create the root element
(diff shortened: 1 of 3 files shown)


spencerseale (approved):
Looks good to me! Thanks Kostas!

tiledb/bioimg/converters/base.py
```diff
@@ -536,7 +544,10 @@ def to_tiledb(
                 original_metadata = reader.original_metadata
             else:
                 if ome_xml := reader.original_metadata.get("ome_metadata"):
-                    pruned_metadata = remove_ome_image_metadata(ome_xml)
+                    if isinstance(exclude_metadata, bool):
+                        pruned_metadata = remove_ome_image_metadata(ome_xml)
+                    else:
```
ihnorton, line 549:
```suggestion
                    elif callable(exclude_metadata):
```
and then raise an exception if it is neither.

tiledb/bioimg/converters/base.py
```diff
@@ -536,7 +544,12 @@ def to_tiledb(
                 original_metadata = reader.original_metadata
             else:
                 if ome_xml := reader.original_metadata.get("ome_metadata"):
-                    pruned_metadata = remove_ome_image_metadata(ome_xml)
+                    if isinstance(exclude_metadata, bool):
+                        pruned_metadata = remove_ome_image_metadata(ome_xml)
+                    elif callable(exclude_metadata):
+                        pruned_metadata = exclude_metadata(ome_xml)
```
sgillies, line 550: @ktsitsi callbacks all tricky. It's almost guaranteed that accidents will happen. I think it would be good to have some extra precise error handling here. Something like this:

```suggestion
                        try:
                            pruned_metadata = exclude_metadata(ome_xml)
                        except Exception as exc:
                            raise MetadataCallbackError("OME XML callback function failed") from exc
```

ktsitsi: Thank you @sgillies. nice catch! I added the changes.

PR #153, opened 2025-02-21
Bioimg package dependencies

ktsitsi:
This PR:

- In the past `openslide-python` required the installation of the `openslide` library, which was system dependent.
- With `openslide-bin` introduction and python3.8+ we avoid any manual setup for the different systems we support (Linux, Darwin and Windows)
- Adds the `openslide-bin` as a project dependency
- `Openslide-python` is now automatically find the installation of the `openslide` binaries.
- Removes deprecated ReadMe.md installation instructions
- CI steps for installing the `openslide` binaries on different setups.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,25 +37,6 @@ jobs:
           pre-commit
           pytest-cov
           pytest-mock
-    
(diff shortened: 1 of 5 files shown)


bkmartinjr (approved):
nice improvement!

PR #155, opened 2025-03-11
Python 3.12 issue resolution and CI enrichment

ktsitsi:
This PR:

- Introduces conditional function calls to `importlib.metadata` library according to the python version.
- Adds a CI job for running CI with `python 3.12` too.


shaunrd0 (approved):
Thanks!

